PR #5, opened 2016-10-24
Assorted updates - example(opq) now prints something

Robinlovelace:


R/opq.R
```diff
@@ -18,6 +19,9 @@
 #' \dontrun{
 #' overpass_query(q) -> reading_noms
 #' sp::plot(reading_noms$osm_nodes)
+#' p <- overpass_query(opq("Leeds") %>%
+#'   add_feature("leisure", "park"))
+#' sp::plot(p$osm_polygons[1,])
```
Robinlovelace: Hi @mpadge I think there may be some debugging to do here, these things fail for me here (wanted users with mapview to see the parks):

```
#' if(require(leaflet)){
#'   mapview::mapview(p$osm_polygons[1,])
#' }
```

Also, why are the parks visible in the first but not second line below?

```
#' sp::plot(p$osm_polygons[1:10,])
sp::plot(p$osm_polygons)
```

Strange.

Robinlovelace: P.s. I think the result for "parks" should definitely be a SpatialPolygonsDataFrame - good use case of when autodetecting of most likely desired output is good - it should decide for you unless specified by an argument, e.g. `output_type = "points"` if the user really wants parks that OSM contributors have erroneously added as points.

mpadge:
Thanks Robin! Great stuff! The only thing I'll undo is the inclusion of the `.Rd` files for the `src/*.cpp` functions, which are only added because `Rcpp` requires the `[[Rcpp::export]]` flag which inadvertently causes `(r)oxygen` to doc those. I'll add them to the ignore lists. Adding `tmap` is a fine idea.

R/opq.R
mpadge: This all has to be considered in relation to [this issue](https://github.com/osmdatar/osmdata/issues/1), the only practicable resolution of which is likely to be the necessity of defining a (simple S3) class for the returned object. Defining a class would also allow us to return, for example, the actual query and a timestamp. As for your probs: Inserting items of defined classes into a list is not really good practise (and is my fault here!), and is likely the reason for your plot issues.

mpadge:
```
require (sp)
sldf <- new ("SpatialLinesDataFrame")
lst <- list (lines1=sldf, lines2=sldf)
lst [[1]]
# gives appropriate output
lst
# $lines1
# Error in ...
```
That's why the `print` method doesn't work properly. You `sp::plot` problem will be for similar reasons

PR #10, opened 2016-11-19
Fix error with printing returned object and a few other changes

Robinlovelace:
Think these are improvements but please review and let me know your thoughts.

Use cases that used to fail:

```r
originsbb = tmap::bb("Sevilla")
q = osmdata::opq(bbox = bb) %>% 
  osmdata::add_feature("place", "*")
origins = osmdata::overpass_query(q)
origins
```

And due to lacking bb:

```r
only_nodes = "[out:xml];\nnode\n  [\"highway\"=\"bus_stop\"]\n  [\"shelter\"]\n  [\"shelter\"!~\"no\"]\n  (50.7,7.1,50.8,7.25);\nout body;"
query = only_nodes
overpass_query(only_nodes)
## Issuing query to OSM Overpass...
## Rate limit: 2
## Error in query$bbox : $ operator is invalid for atomic vectors
```

We don't need to force the user to supply a bb if `query` is a char string - that's an overpass issue and if the text string is garbled in any way it will tell the user. So also only enforced having a `bb` if query is a list.


man/osmdata.Rd
```diff
@@ -1,11 +1,26 @@
 % Generated by roxygen2: do not edit by hand
-% Please edit documentation in R/osmdata-package.R
+% Please edit documentation in R/osmdata-class.R, R/osmdata-package.R
 \docType{package}
 \name{osmdata}
 \alias{osmdata}
 \alias{osmdata-package}
-\title{Tools to Work with the OpenStreetMap (OSM) Overpass API}
+\title{Create osmdata object}
```
Robinlovelace: Woops - wasn't intending to overwrite that!

mpadge:
Thanks @Robinlovelace - looks grand, and ready to pull ... after you fixed the couple of things that break on travis. `base_url` needs a doc entry, but the other errors are because you've `@export`ed the `osmdata` ''funnction''. This is, however, just a class constructor which i can't see any reason to export. The only real reason to export constructors is to enable creation of `new()` ones, but as far as I see it, there won't ever be a need to create an empty object; this function should only ever be called in conjunction with an actual `overpass` query, and the constructor can remain unexposed/un`export`ed. Happy to discuss if you've got a different opinion and can see a reason to `export` the constructor, otherwise, just ditch the `@export` there and travis should then pass. You can also check your own builds by simply switching on your own fork within Travis. Thanks!

Robinlovelace:
Sounds good - yes no need to export `osmdata()` - was just frustrating calling it. How about referring to it as `overpass:::osmdata()` when used in function calls? Happy with the rest of the review, on it!

Robinlovelace:
Sounds good - yes no need to export `osmdata()` - was just frustrating calling it. How about referring to it as `overpass:::osmdata()` when used in function calls? Happy with the rest of the review, on it!

Robinlovelace:
Sweet it passed! Heads up @mpadge

PR #26, opened 2016-12-26
Getbb test/mock

Robinlovelace:


Robinlovelace:
Not sure if this is even useful - couldn't get my head round **vcr** sufficiently, as you'll see...

mpadge:
Looking forward to checking this out myself - it certainly makes for beautifully simple code. I'll get back to in a few days on that one. Appreciate the efforts!

Robinlovelace:
No worries - just a matter of figuring out `vcr` probably and happy if this PR gets rejected as it's very rudimentary... One thought though: suspect it will make our lives easier if we wait until they're on CRAN no? At least an indication of stability... All stuff to chat about in NY.

PR #40, opened 2017-02-25
Review vignette

Robinlovelace:


vignettes/osmdata.Rmd
```diff
@@ -49,11 +49,11 @@ An `osmdata` query consists of an initial `opq()` query and one or more features
 q <- opq (bbox="greater london uk")
 q <- add_feature (q, key="highway", value="motorway")
 ```
-The corresponding data may then be extracted as `R Spatial` objects with,
+The corresponding data may then be extracted as a `Spatial` object defined by the `sp` package, with,
 ```{r osmdata_sp}
 osmdata_sp (q)
 ```
-or an `R Sinple Features` objects with
+or an `R Simple Features` objects with
```
Robinlovelace: @mpadge I think it would be better to show the user first how to download the raw data onto the hd before converting it - more bandwidth efficient and a chance to show the `request -> dl -> conversion` steps in process. Happy to do this but cannot remember the function to simple dl raw OSM data.

Robinlovelace: That way the data is useful regardless of if you use R for analysis or not, as shown here:

osm-qgis

mpadge:
Thanks Robin! Could you please just make one modification: `osmdata _xml()` has a file argument to enable direct downloading without needing your `writeLines` example. Files can also be passed directly to all `osmdata...()` functions.

Robinlovelace:
Thanks @mpadge - I've fixed that issue and removed the duplication (sorry, should have read ahead!).

Still more to go. I think making `dat_sp` and `dat_sf` separate is worthwhile. Do you want to iterate on this version (check out the QGIS image) or shall I carry on? I slightly prefer the former option as it takes the pressure off me + allows quick iteration.

PR #42, opened 2017-02-26
Completing vignette

Robinlovelace:
This mentions previous packages `osmar` and `overpass`, has some references now, and demonstrates `qry_to_string()`, among other things.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,5 +19,5 @@ man/rcpp_test_polygons.Rd
 # vim files
 .*.un~
 .*.swp
-# osmdata files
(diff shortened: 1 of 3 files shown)


Robinlovelace:
Job done, aside from stylistic issues ; )

Can I remove the non-standard use of whitespace before brackets and add spaces between arguments please? If so I promise to stop using `=` for assignment on this project!

In any case, almost over to @masalmon - I think you'll like the cycle path example!

mpadge:
yeah, permission granted to remove whitespace before brackets

PR #62, opened 2017-05-18
Improve error handling of getbb

Robinlovelace:


mpadge:
thanks heaps @Robinlovelace for the quick rescue! One travis build failed, but it's also being a bit flaky this afternoon, so i just re-started it. It's really bloody hot here in Austria/Germany, and both travis (Berlin) and nominatim (Germany somewhere I think?) are having issues ... interesting.

Robinlovelace:
I think nominatim is working fine - it just doesn't like being hammered by a for loop would be my interpretation.

mpadge:
yeah, but it was failing heaps often enough for the last attempted CRAN submission to get rejected straight away - failed on 2 of the 1st 3 machines tested. No examples or tests actually call nominatim now, so it should be okay

Robinlovelace:
Many thanks for the sterling (or should that be stubbing?) work sir!

Issue #68, opened 2017-06-20
Custom API


fzenoni:
Is there a way to select a custom Overpass API, similarly to what is possible in osmar with the   `osmsource_api(url = "http://api.openstreetmap.org/api/0.6/")` function?

mpadge:
Um, whoops - thanks for pointing that out! That functionality is kinda there, but it's buried in the now non-exported function [`overpass-query()`](https://github.com/ropensci/osmdata/blob/master/R/overpass-query.R#L90). This got switched somewhere along the line to default-only use of this value. I'll just have to re-insert this argument to the [`osmdata_xml/sp/sf()` functions](https://github.com/ropensci/osmdata/blob/master/R/get-osmdata.R), add appropriate documentation, and pass on through to `overpass_query()` and it'll work again.

As previously: PRs welcome if you want to give this a go

fzenoni:
I've never contributed to any package before, so I guess there's always a first time :)
I will try, and let you know if I fail.

mpadge:
Actually going to keep this issue open until this ability is described in the vignette

fzenoni:
Maybe this isn't the end of the story. When I install the package from github and R tries to load the help box for any given function, I get this:
```
Error in fetch(key) : 
  lazy-load database 'C:/Users/fzenoni/Documents/R/win-library/3.4/osmdata/help/osmdata.rdb' is corrupt
```

mpadge:
Not sure what that could be - maybe just a conflict as I was pushing to github or something? If it persists, then feel free to open another issue with full code to reproduce the problem.

fzenoni:
My anxiety was unnecessary: reinstalling the package solved the issue. Thanks again.

Issue #59, opened 2017-04-13
clarify distinction with OpenStreetMap package


mpadge:
This should be done, notably in response to @edzer's [ropensci review of the package](https://github.com/ropensci/onboarding/issues/103). We just need to work out the best place / places to do it. cc @Robinlovelace @maelle

maelle:
README + intro vignette?

Robinlovelace:
Yes to @maelle's suggestions - any easy fix.

mpadge:
All agreed then. I'll happily volunteer for [the harder issue](https://github.com/osmdatar/osmdata/issues/60), so it'd be great if one of you could take care of that

maelle:
I can but from next week :-)

mpadge:
no rush: i won't be able to anything more until at least the week after anyway. plus it seems we've still got another two reviews coming.

maelle:
only one other review, right?

PR #84, opened 2017-07-12
bb fixes

Robinlovelace:
Look like we've both worked on this : (


Robinlovelace:
But I fixed the merge conflict with meld : )

mpadge:
That looks awesome! Thanks so much for the labours. Just one thing: The bit in the vignette about "different overpass API servers" (around previous line#164) is actually really important, and was implemented via #68. Did you mean to delete that?

Robinlovelace:
Yes good point - will do.

mpadge:
@Robinlovelace one more thing while your polishing the vignette: do you think you could address #59? That would be awesome!

PR #106, opened 2017-11-25
Fix #104

### LINKED ISSUE #104: Allow getbb to return an sf_polygon

Robinlovelace:


Robinlovelace:
I've not rebuilt the documentation or implemented the case for multipolygons which should be easy but would prefer it if you did that your end if poss pls 👍

Robinlovelace:
Reproducible example for Leeds:

```r
library(tmap)
tmap_mode("view")
bb_sf = getbb("leeds", format_out = "sf_polygon")
nrow(bb_sf)
qtm(bb_sf[1,])
```

image

mpadge:
grand stuff. I'll just have to replace the `sf` code with the `osmdata` equivalent to ditch the direct `sf` calls. I'll try to paste the necessaries here tomorrow. If not, tues latest. Thanks!

Robinlovelace:
How come you want to ditch the sf calls in there?

mpadge:
Here's the non-sf way to convert the matrix to an `sf` object, because `osmdata` still functions without direct `sf` dep. One further important modification: Our code currently auto-selects the first polygon object, but we could and should insert an extra parameter in `getbb` called `poly_num` (or something) to allow selecting other polygons. Maybe with `silent = FALSE`, we should also output the number of polygons returned.

Non-sf dep then requires the following additional function in `getbb.R` (noting that `make_sf` is an internal `osmdata` function):
```
mat2sf <- function (mat, pname)
{
    mat_sf <- list (mat)
    class (mat_sf) <- c ("XY", "POLYGON", "sfg")
    mat_sf <- list (mat_sf)
    attr (mat_sf, "class") <- c ("sfc_POLYGON", "sfc")
    attr (mat_sf, "precision") <- 0
    bb <- as.vector (t (apply (mat, 2, range)))
    names (bb) <- c ("xmin", "ymin", "xmax", "ymax")
    class (bb) <- "bbox"
    attr (mat_sf, "bbox") <- bb
    crs <- list (epsg = 4326L,
                 proj4string = "+proj=longlat +datum=WGS84 +no_defs")
    class (crs) <- "crs"
    attr (mat_sf, "crs") <- crs
    attr (mat_sf, "n_empty") <- 0L
    mat_sf <- make_sf (mat_sf)
    names (mat_sf) <- pname
    attr (mat_sf, "sf_column") <- pname
    return (mat_sf)
}
```

Then your code in current PR:
```
if(format_out == "sf_polygon") {
+      ret = lapply(X = ret, function(x) {
+        if(is(x, "list")) {
+          x <- x[[1]]
+        }
+        sf::st_polygon(list(x)) 
+      }) %>% 
+        sf::st_sfc() %>% 
+        sf::st_sf() %>% 
+        sf::st_set_crs(4326)
+    }
```
can be replaced with:
```
if(format_out == "sf_polygon") {
+      ret = lapply(X = ret, function(x) {
+        if(is(x, "list")) {
+          x <- x[[poly_num]]
+        }
+        mat2sf (x, pname = place_name)
       })
+    }
```
Note that i'm not sure what your `lapply` is doing here? If we have a `poly_num` arg, then there is no `list`, just a single polygon. That's probably the best approach because it allows for much safer piping. Otherwise type consistency would be broken because `getbb(..,format_out="sf_polygon")` would return an `sf` object if `nominatim` happened to deliver a single polygon, or a list of objects is there were multiples. The class of the result could never be predicted in advance, which seems unwise.

Finally, note a singular advantage of this over `sf`: The object ends up with the geometry having the `place_name`, which is pretty darn useful!

Robinlovelace:
Nudge - good to go I think @mpadge

PR #149, opened 2018-10-25
on.load picks a new API each time

JimShady:
I've tried to add the API sample as discussed @mpadge . My first pull request so fingers crossed ! :-)


mmd-osm:
Note that `http://overpass-api.de/api/interpreter` is in fact an alias for `lz4.overpass-api.de` and `z.overpass-api.de`. So total there are only two servers, hence  I would recommend to remove lz4.* and z.* form the list and let DNS round robin do its job.

Issue #180, opened 2019-07-29
bbox queries without features does not work


neogeomat:
Running this code 
```
q <- opq ("hampi india")
opq_string (q)
```
gives `"[out:xml][timeout:25];\n(\n);\n(._;>;);\nout body;"` which does not contain bbox parameter.  so queries with bbox but without features don't work.

PR #181, opened 2019-07-29
bbox parameter added to query when features is not given

neogeomat:


mpadge:
Great work, and a pretty important improvement. I had originally intentionally coded the package **not** to accept `opq() %>% osmdata_s*()` without `add_osm_feature()` in the middle, because doing that is really an abuse of the overpass server. But this is not clearly stated anywhere, and the way it current fails (your issue #180) is totally unhelpful. This PR is a great solution, but just needs a couple of modifications. Could you please:

1. Make sure it works properly by including curly brackets so your current `res <- ` with the `bbox` values included comes inside the curly brackets
2. Put the current `res <-` as the corresponding `else` condition.

If you want, you could also pass the `quiet` parameter though to `opq_string()` (because that function is only called directly from the various `osmdata_s*()` functions which all have a `quiet` parameter) and issue an informative message inside `if (is.null (res)){...}` advising against queries without `add_osm_feature()`.

Issue #186, opened 2019-09-18
Doubts reading the vignettes


agila5:
Hi, today I read the vignettes of thepackage and run the code but I still have some doubts. 

1.  When I run the code in the section 2.1 where you explain how to combine several features of OSM I get different results then yours. This is not surprising since you probably run those lines several months ago but there is still something that I cannot understand. This is the code: 

``` r
# packages
library(osmdata)
#> Data (c) OpenStreetMap contributors, ODbL 1.0. http://www.openstreetmap.org/copyright

# data
q <- opq(bbox = 'Kunming, China') %>%
  add_osm_feature(key = 'natural', value = 'water') %>%
  add_osm_feature(key = 'name:en', value = 'Dian', value_exact = FALSE)

dat1 <- opq(bbox = 'Kunming, China') %>%
  add_osm_feature(key = 'natural', value = 'water') %>%
  osmdata_sf ()
dat2 <- opq(bbox = 'Kunming, China') %>%
  add_osm_feature(key = 'name:en', value = 'Dian', value_exact = FALSE) %>%
  osmdata_sf ()
dat <- c (dat1, dat2)

unlist (lapply (osmdata_sf (q), nrow) [4:8])
#>         osm_lines      osm_polygons osm_multipolygons 
#>                 2                 1                 1
unlist (lapply (dat1, nrow) [4:8])
#>        osm_points         osm_lines      osm_polygons osm_multipolygons 
#>             49055               175               831                44
unlist (lapply (dat2, nrow) [4:8])
#>        osm_points         osm_lines      osm_polygons    osm_multilines 
#>              4131                44                 4                 1 
#> osm_multipolygons 
#>                 1
unlist (lapply (dat, nrow) [4:8])
#>        osm_points         osm_lines      osm_polygons osm_multipolygons 
#>             51112               217               834                44
```

<sup>Created on 2019-09-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

There are two things that are not clear to me: a) how is it possible that when you add the two features with a logical AND (i.e. the `q` object) then the result is formed by 4 objects and no osm_point? Probably I don't get something because I was thinking that every point object stored in osm_points was also a point of the others lines/polygons (as you explain in section 3.2) so I don't see how the two features share a line and several polygons but no point; b) what happened to the osm_multiline object of `dat1` when you merged `dat1` and `dat2` with `c(dat1, dat2)`? Because when I print the output of `dat` I see no osm_multiline. 

[...]

PR #201, opened 2020-01-05
Update AND and OR examples

agila5:
Hi! I update the examples in the vignette according to your explanation in #186. 

In particular, when you define `dat <- c (53777, 238, 908, 1, 49)` I think you should set the number of multilinestrings equal to 1 instead of 0 and exclude the multilinestring output just from `dat1`, i.e. `dat1 <- dat1 [c (1:3, 5)]` instead of `dat <- dat [c (1:3, 5)]`. 

The other change is used to automatically update the description of the examples using R code instead of fixed numeric input. I'm not sure if it works on the online vignette or if you like that so I changed that only in two places but I could provide more commits if you thinks it's a good idea.

### DIFF
--- a/vignettes/osmdata.Rmd
+++ b/vignettes/osmdata.Rmd
@@ -414,7 +414,7 @@ nrow(des_bike_and_bridge$osm_points); nrow(des_bike_and_bridge$osm_lines)
 ## [1] 32
 ```
 
-That query returns only 7 points and 3 lines.  Designed cycleways **OR** bridges
+That query returns only 99 points and 32 lines.  Designed cycleways **OR** bridges
 can be obtained through simply combining multiple `osmdata` objects with the `c`
 operator:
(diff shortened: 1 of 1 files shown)


mpadge:
That's great Andrea, thanks very much for the input. It is currently failing on travis because when you reduce `dat1` to a length of 4 via
``` r
dat1 <- dat1 [c (1:3, 5)]
```
then the subsequent [lines#301-306](https://github.com/ropensci/osmdata/blob/master/vignettes/osmdata.Rmd#L301-L306) fail because all five names are re-assigned to these objects. An easy fix would be to simply screen-dump
``` r
dat1 [c (1:3, 5)]
```
leaving the actual object with all five items. Once you modify that and travis passes, then I'll happily merge

agila5:
Should be good to go!

Issue #212, opened 2020-02-15
extend available_tags/available_features to use taginfo service


espinielli:
See proposal from @mpadge in #207 for an extension of `available_tags` and `available_features`

mpadge:
@espinielli Can you please confirm that both this issue, and the associated PR #213 can be closed now? (or even better, just close them if so). Thanks!

espinielli:
Sorry for late reply. yes I confirm.

PR #210, opened 2020-02-15
Fix available_features and available_tags

espinielli:
Fixes #211 by extracting from `href` rather than `title`.
Title entries in fact contain spaces in lieu of underscores.
Fixes #207

PR #213, opened 2020-02-16
available functions optionally use taginfo

espinielli:
This is a PR for #212


mpadge:
@espinielli This one can now be closed, because actually all done via #210, right?

PR #228, opened 2020-10-06
fix code style for unname_osmdata_sf

espinielli:
I was reading the documentation for `unname_osmdata_sf` and spotted a small issue with documentation, i.e. missing initial \` when refering to `osmdata`


Robinlovelace (approved):
Definite fix I think :+1: Will allow @mpadge to merge though.

Issue #229, opened 2020-11-10
[BUG] osm_enclosing misses lat lon in string


idshklein:
I'm trying to use the new osm_encolsing function, and observed that the example from the documentation does not manage to download the data. when I observed more carefully, I found the the overpass api query doesn't paste the lat and lon of the requested point.

``` r
library(osmdata)
#> Warning: package 'osmdata' was built under R version 4.0.3
#> Data (c) OpenStreetMap contributors, ODbL 1.0. https://www.openstreetmap.org/copyright
lat <- 54.33601
lon <- -3.07677
key <- "natural"
value <- "water"
enclosing = "relation"
timeout = 25
opq_enclosing(lon, lat, key, value) %>%
  opq_string()
#> [1] "[out:xml][timeout:25];\n(\nis_in(,)->.a;relation(pivot.a) [\"natural\"=\"water\"];);\n(._;>;);\nout;"
```

<sup>Created on 2020-11-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> - Session info ---------------------------------------------------------------
#>  setting  value                       
#>  version  R version 4.0.2 (2020-06-22)
#>  os       Windows 7 x64 SP 1          
#>  system   x86_64, mingw32             
#>  ui       RTerm                       
#>  language (EN)                        
#>  collate  English_United Kingdom.1252 
#>  ctype    English_United Kingdom.1252 
#>  tz       Asia/Jerusalem              
#>  date     2020-11-10                  
#> 
#> - Packages -------------------------------------------------------------------
#>  package     * version date       lib source        
#>  assertthat    0.2.1   2019-03-21 [1] CRAN (R 4.0.2)
#>  backports     1.2.0   2020-11-02 [1] CRAN (R 4.0.3)
#>  callr         3.5.1   2020-10-13 [1] CRAN (R 4.0.3)
#>  cli           2.1.0   2020-10-12 [1] CRAN (R 4.0.3)
#>  crayon        1.3.4   2017-09-16 [1] CRAN (R 4.0.2)
#>  curl          4.3     2019-12-02 [1] CRAN (R 4.0.2)
#>  desc          1.2.0   2018-05-01 [1] CRAN (R 4.0.2)
#>  devtools      2.3.2   2020-09-18 [1] CRAN (R 4.0.3)
[...]

PR #230, opened 2020-11-10
Fixes ropensci/osmdata#229

idshklein:


mpadge:
Thanks, that's almost it, but now you need an opening and closing curly brace - see the failing test results in any of the R CMD check links above.

idshklein:
I see this is not working....
I think the query has more problems. playing with it a bit in overpass turbo to see what other fixes should be made.

Issue #257, opened 2021-12-04
[FEATURE] Limit queries to specific feature types


Mashin6:
Implementation of this is dependent on #255.

Currently each query is run on all three types of objects `node`, `way` and `relation` written as three separate lines. There is only one way how to change this and that is with `opq(..., nodes_only = TRUE)` that sets an attribute in opq object. If #255 gets implemented I could imagine adding another parameter that would specify type of objects.

`get_osm_features(match_all = c("natural = tree", "leaf_type = broadleaved"), type = "n")`

Possible `type` values and their translations:
`n`: `node[natural=tree][leaf_type=boradleaved];`
`w`: `way[...`
`r`: `rel[...`
`nw`: `nw[...`
`nr`: `nr[...`
`wr`: `wr[...`
`nwr`: `nwr[...`
 
This will be stored by `get_osm_features` in new `opq$type` slot. Length of `$type` has to match length of `$features`.

mpadge:
Addressing this would also require finally thinking about the by-now-four-year-old #55. In principle at the overpass side, each of these combinatorial possibilities (so from `nw` down) offers the possibility of recurse down. They all offer the possibility of recurse up, as well as simply queries with no recursion. So implementing even #55, let alone this one, opens up a truly bewildering array of possibilities in place of current form which offers no such options but a compelling simplicity in place of this.

This simplicity was a very large part of the initial motivation for the package, and not something we would want to compromise. That just means that, if this were to be implemented, it would need to be done in a way that was as nonintrusive as possible, so nobody who didn't care about this would have to.

And now the even bigger issue: The current "always-recurse-down" implementation ensures that all queries return sensibly complete spatial objects. Embarking upon any of these options opens up the possibilities of returning objects that make sense in the OSM domain, but which are difficult to convert into Simple Features format. I personally don't have any great issue with that, and would love a world in which everybody was happy with `sc` instead of `sf`, but that's not the world we have, and I'm sure the vast majority of `osmdata` users use `osmdata_sf`. You can't covert a raw relation into a spatial object, and without recursion, it's equally difficult to convert many of the other possibilities. All just means that, like #255, this needs some very careful thought and design before proceeding. But thanks for starting the discussion, and it will be interesting to see where these considerations take us.

Mashin6:
I think it is totally possible to keep it simple with good defaults.

[...]

Issue #256, opened 2021-12-04
[FEATURE] Data trimming on OP server side


Mashin6:
As suggested in #178 trimming by area polygon using Overpass is fast and useful way of data filtering. 

_Note: OP now treats all [closed ways as areas](http://dev.overpass-api.de/blog/way_based_areas.html) thus allowing filtering by features like parks or buildings.
Overpass-turbo uses {{geocodeArea:"name"}} shorthand to query Nominatim and takes the first retrieved object for area filtering. We could use this behavior here as well._

Ok. I took a look on the existing code and though of a solution. But before I write any code I want to check if the solution is in line with overall spirit of the package structure and any long term visions:

1. `getbb(... , format_out = "data.frame")` will be used to query Nominatim for geocoded area.
2. Resulting data.frame can be then passed into new function `overpass_trim(dat, osm_area)` in a similar way as it is used for `trim_osmdata()` and `format_out = "polygon"`.
`osm_area =` data.frame from getbb()
3. `overpass_trim()` will then take only the very first result of Nominatim output for filtering. (maybe show warning if multiple results?)
4. `overpass_trim()` can also take two other parameters `osm_id` and  `type` that can be used to trim by a specific osm object instead of relying on Nominatim search.
`osm_id =` single id or vector of ids
`type =` single type or vector of types (w|r)
5. `overpass_trim` will parse and set `opq$bbox=NULL` and new `$trim_area= list( c(id_1, id_2), c("r", "w"))`
6.  `opq_string_intern()` will detect presence of $trim_area in opq object and modify query buildup.
So from this:
``` R
opq() |>
    trim_osmdata(osm_id = c(id_1, id_2), type = c("r", "w")) |>
    add_osm_feature(key = "natural", value = "tree") |>
    osmdata_sf()
```
we get
```
(
rel(id:id_1);
way(id:id_2);
);

map_to_area->.a;

(
nwr[natural=tree](area.a);
);
```
It would work the same with `getbb` nominatim search but the `osm_id` and `type` will be taken from the passed data.frame.

jmaspons:
Can be closed as per #286

mpadge:
Then please do so @jmaspons, so you'll be recorded as the having done the work required to close this. Thanks!

jmaspons:
> Then please do so @jmaspons, so you'll be recorded as the having done the work required to close this. Thanks!

Can I? I think I have no perms

PR #258, opened 2021-12-11
Adding overpass_trim() for filtering by area in overpass

Mashin6:
This PR addresses suggestion #256.

Main changes are:
1. New feature added `overpass_trim()` that modifies OP query to subset OSM features by area rather than currently default bbox. Function parses entered information about areas to be used for trimming and stores them in a list in `overpass_query` object: `trim_area = list (ways = c(ID, ... ), relations = c(ID, ...)`
2. `opq_string_intern()` was modified to accommodate changes necessary for area filtering.
3. A new variable for feature type `ftype` was introduced in `opq_string_intern` that allows easy switching of which object types are queried. This can be easily extended to more types when #257 would be implemented.
4. OP queries for nodes, ways, relations are not as 3 separate rows, but as a single `nwr[...];`
5. WARNING: The most controversial change could be that I removed checks for `if (is.null (bbox))` from functions: `opq()`, `add_osm_feature()`, `add_osm_features()` and introduced final check at the beginning of `opq_string_intern`. This was to allow `opq()` to be called without `getbb()`. So the queries can be now built like this:

``` R
opq() |>
        add_osm_feature(key = "natural",
                        value = "tree") |>
        overpass_trim(id = c(11597767, 43437030), 
                      type = c("relation", "way") ) |>
        osmdata_sf()
```
or
``` R
a <- getbb("portsmouth usa", format_out = "data.frame")
opq() |>
        add_osm_feature(key = "amenity",
                        value = "restaurant") |>
        overpass_trim(osm_area = a) |>
        osmdata_sf()
```

---
The only thing I am not happy about is that a typical user would expect trimming to work like this instead:
```
opq("portsmouth usa") |>
        add_osm_feature(key = "amenity",
                        value = "restaurant") |>
        overpass_trim() |>
        osmdata_sf()
```
But that would require making `bbox_to_string` return object id and type along with its bbox and let `opq` sort it out into `$bbox` and `$trim_area`.  Any ideas?


jmaspons:
Are you still open to work on this? I'm interested in the functionality and I can help in the implementation. Any feedback from the contributors would be useful.

mpadge:
@jmaspons yes, i am very keen for something to happen with this - it's already been open way too long, which ain't good :scream: I should also be able to allocate some time to this soon too, but please feel free to see what you can do in the meantime. It's an important idea.

Mashin6:
@mpadge Would you be able to provide any feedback on the code/changes proposed? There were deeper modifications in the underlaying code and would be good to know your opinion if this is even in the right direction.

mpadge:
Yeah, for sure @Mashin6, and great that you responded so promptly after such a long time - thank you! I'll get back to asap - hopefully in a  few days when i find some time to look through it. Thanks!

Mashin6:
@jmaspons Did you have some ideas for implementing the overpass-side area trimming?

jmaspons:
My idea is to pass a string in the form `rel(id)` to `bbox` parameter, which I would rename to `area` at some point. I have a draft https://github.com/jmaspons/osmdata/commit/cceab6f1a046a59f4b40f75cdd61e7752998ff75

Then, a `get_area` function, similar to `getbb` for areas, would help to get the string. If you like the idea, I can test the code and implement the `get_area` function and all de details.

mpadge:
@Mashin6 There've been a few changes to `opq.R` recently. Would you please be able to merge those within your branch, and then push the updated version to this PR? Thanks!

jmaspons:
> I like the simplicity of it. Would definitely like to see how it turns out.

You can find a working implementation at https://github.com/ropensci/osmdata/pull/286

Issue #254, opened 2021-11-30
Autosplit large queries


Mashin6:
Currently if query times out or runs out of memory the request is terminated. Given that the amount of data in OSM is growing it might be worth thinking about how to help user in these situations. 
A possibility is that if a query fails, the bbox is split in the middle into four equal rectangles and each are submitted one after another. If any of them fails then it is recursively split until the correct size is reached when the query successfully retrieves data.

Similarly as here:
https://github.com/ZeLonewolf/osm-overpass-scripts/blob/7bf5ff10948f50ee5f1e33b08a9d0852274094fb/get_tag_density_map.sh#L139

Notes:
* Might be good to ask user before proceeding with bbox splitting
* Needs a way how to handle possible duplicates retrieved at the bbox borders.

mpadge:
Thanks @Mashin6, I've actually got code elsewhere that does exactly that. The handling of duplicates is already in-built via the `c` operator; you just have to expand each bbox part so that there is some non-zero overlap, then combining all with `c` will automatically de-duplicate everything. The real issue here (in my opinion at least) is that enabling that would extend an open invitation for people to really start abusing the overpass server. And so: my interim suggestion would be, at least initially, to add an extra vignette on the general procedure. Would you be interested in dong that?

Mashin6:
I've never wrote a vignette, but I could give it a try.

By general procedure you mean how to split bbox, run several separate queries and then manually merge the results into one?

A solution to the extensive server load could be restricting this type of querying to kumi OP only. They don't really have any restrictions about usage.
_As a warning, kumi has broken attic data so it is not good for using queries that contain `datetime =` or `datetime2 =`_

mpadge:
@Mashin6 yes, a vignette on splitting a bbox, running separate queries, and merging results is exactly what i mean. I've started a generic vignette template in #262 - feel free to extend however you like. If you do contribute, please make sure you add your name both to the `author` field of that vignette and to the `DESCRIPTION` file as `ctb`. Thanks!

jmaspons:
Is this issue done?

PR #264, opened 2022-01-09
Extending query split vignette

Mashin6:
Adding examples of bbox splitting for larger queries.
#254

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -11,6 +11,7 @@ Authors@R: c(
     person("Andrea", "Gilardi", role = "ctb"),
     person("Enrico", "Spinielli", role = "ctb"),
     person("Anthony", "North", role = "ctb"),
+    person("Martin", "Machyna", role = "ctb"),
     person("Marcin", "Kalicinski", role = c("ctb", "cph"),
            comment = "Author of included RapidXML code"),
     person("Finkelstein", "Noam", role = c("ctb", "cph"),
--- a/vignettes/query-split.Rmd
+++ b/vignettes/query-split.Rmd
@@ -2,6 +2,7 @@
 title: "4. Splitting large queries"
 author: 
   - "Mark Padgham"
+  - "Martin Machyna"
 date: "`r Sys.Date()`"
 bibliography: osmdata-refs.bib
 output: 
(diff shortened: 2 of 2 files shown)


mpadge:
Thanks @Mashin6, that looks great. One immediate thought is that the sub-divided parts of a bbox really need to have a small degree of overlap to ensure that all objects that cross the boundary lines are appropriately captured. I like the way you've coded the bbox division - would it be possible to extend that elegant simplicity to include an additional parameter to control proportional overlap across boundaries? I usually use a default overlap of 5%, and once had code somewhere that showed that it did make a difference. (That was, however, from an older version of overpass, so maybe no longer valid?)

Mashin6:
Yeah, I guess one can add a fudge factor that would extend the newly made boundaries by a bit.
```R
split_bbox <- function(bbox, grid = 2, ff = 0.05) {
    ...
    b <- matrix(c(xmin + ((i-1) * dx),
                  ymin + ((j-1) * dy),
                  xmin + (i * dx) + (dx * ff),
                  ymin + (j * dy) + (dx * ff)),
                nrow = 2,
                dimnames = dimnames(bbox))
    ...
}
```

But we should not extend past the original bbox, so there must be a switch for the last edge. e.g.
```R
split_bbox <- function(bbox, grid = 2, ff = 0.05) {
    ...
    b <- matrix(c(xmin + ((i-1) * dx),
                  ymin + ((j-1) * dy),
                  xmin + (i * dx) + (dx * ff * (if (i == grid) 0 else 1)),
                  ymin + (j * dy) + (dx * ff * (if (j == grid) 0 else 1))),
                nrow = 2,
                dimnames = dimnames(bbox))
    ...
}
```

Though I don't know how much it matters. Overpass seem to return also objects that are directly positioned on the bbox edge. For example this [node](https://www.openstreetmap.org/node/2789765493) has coordinates `41.3084017, -72.9262955` and is returned by this query `node(41.3084017, -72.9262955, 41.3085, -72.925);`. OSM works with 7 decimal places and coordinates with more precision get rounded. So potential problem could be when the new bbox after splitting has coordinates with more than 7 decimal places and it gets rounded in the wrong direction. Then at least in theory it should be sufficient to always round up the 7th digit of the x and y max coordinates like:
```R
    b <- matrix(c(xmin + ((i-1) * dx),
                  ymin + ((j-1) * dy),
                  ceiling( (xmin + (i * dx)) * 10^7) / 10^7,
                  ceiling( (ymin + (j * dy)) * 10^7) / 10^7),
                nrow = 2,
                dimnames = dimnames(bbox))
```

Issue #268, opened 2022-03-10
[BUG] Regression (?) in `opq_osm_id()`


agila5:
Hi @mpadge! I want to report a possible regression in `opq_osm_id`. The following example (taken from `?opq_osm_id`) fails with the current CRAN version:

``` r
remotes::install_cran("osmdata", quiet = TRUE)
library(osmdata)
#> Data (c) OpenStreetMap contributors, ODbL 1.0. https://www.openstreetmap.org/copyright
id <- c (1489221200, 1489221321, 1489221491)
opq_osm_id (type = "node", id = id) %>%
  opq_string () %>%
  osmdata_sf ()
#> Error in opq_osm_id(type = "node", id = id): Only a single id may be entered.
```

<sup>Created on 2022-03-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> - Session info ---------------------------------------------------------------
#>  setting  value
#>  version  R version 4.1.2 (2021-11-01)
#>  os       Windows 10 x64 (build 19042)
#>  system   x86_64, mingw32
#>  ui       RTerm
#>  language (EN)
#>  collate  English_United Kingdom.1252
#>  ctype    English_United Kingdom.1252
#>  tz       Europe/Berlin
#>  date     2022-03-10
#>  pandoc   2.16.2 @ C:/PROGRA~1/Pandoc/ (via rmarkdown)
#> 
#> - Packages -------------------------------------------------------------------
#>  package     * version date (UTC) lib source
#>  backports     1.4.1   2021-12-13 [1] CRAN (R 4.1.2)
#>  cli           3.2.0   2022-02-14 [1] CRAN (R 4.1.2)
#>  crayon        1.5.0   2022-02-14 [1] CRAN (R 4.1.2)
#>  curl          4.3.2   2021-06-23 [1] CRAN (R 4.1.1)
#>  digest        0.6.29  2021-12-01 [1] CRAN (R 4.1.2)
#>  ellipsis      0.3.2   2021-04-29 [1] CRAN (R 4.1.1)
#>  evaluate      0.14    2019-05-28 [1] CRAN (R 4.1.1)
#>  fansi         1.0.2   2022-01-14 [1] CRAN (R 4.1.2)
#>  fastmap       1.1.0   2021-01-25 [1] CRAN (R 4.1.1)
[...]

PR #282, opened 2022-11-25
Fix opq_osm_id for more than 1 id

jmaspons:
Fix https://github.com/ropensci/osmdata/issues/268

Examples on the documentation now works


mpadge:
Thanks @jmaspons, that is indeed what  is needed to fix that. Before merging, could you please also update the corresponding tests?

https://github.com/ropensci/osmdata/blob/39d7994ca202b18238782612402fa58a7c59a01f/tests/testthat/test-opq.R#L132-L135

That no longer triggers an error, so you could just change `expect_error` to `expect_s3_class`, with an expected value of "overpass_query", like the example a couple of tests further. Once that's done, I'll be happy to merge. Thanks!

PR #283, opened 2022-11-25
Allow vectors of ids + types for opq_osm_id

jmaspons:
Requires https://github.com/ropensci/osmdata/pull/282

### DIFF
--- a/R/opq.R
+++ b/R/opq.R
@@ -803,7 +803,13 @@ opq_string_intern <- function (opq, quiet = TRUE) {
 
     } else if (!is.null (opq$id)) { # opq with opq_osm_id
 
-        id <- paste(sprintf (" %s(id:%s);\n", opq$id$type, opq$id$id), collapse="")
+        typeId <- data.frame(type=opq$id$type, id=opq$id$id)
+        typeId <- split(typeId, typeId$type)
+        id <- mapply(function(type, ids){
+          paste0(" ", type, "(id:", paste(ids, collapse=","), ");\n")
+        }, type=names(typeId), ids=typeId)
+
+        id <- paste(id, collapse="")
         res <- paste0 (opq$prefix, id, opq$suffix)
 
     } else { # straight opq with neither features nor ID specified
--- a/R/opq.R
+++ b/R/opq.R
@@ -502,7 +502,7 @@ check_features <- function (features) {
 #' must be entered as either a character or *numeric* value (because R does not
 #' support long-form integers). id can also be a character string prefixed with
 #' the id type, e.g. "relation/11158003"
-#' @param type Types of objects (recycled); must be either `node`, `way`, or `relation`.
+#' @param type Type of objects (recycled); must be either `node`, `way`, or `relation`.
 #'   Optional if id is prefixed with the type.
 #' @param open_url If `TRUE`, open the OSM page of the specified object in web
 #' browser. Multiple objects (`id` values) will be opened in multiple pages.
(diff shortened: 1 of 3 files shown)


mpadge:
Thanks @jmaspons, once you've updated tests for #282, i'll merge that and get straight on to this one.

mpadge:
@jmaspons I see what you're trying to do here, and think that this also is a good improvement. It still needs a bit of work though. Overpass queries should have all of the `type` entries grouped together, whereas your current line here:
``` r
id <- paste(sprintf (" %s(id:%s);\n", opq$id$type, opq$id$id), collapse="")
```
Just puts every ID with its own distinct `type`, so submits queries like `(node(id:1); node(id:2); node(id:3))`. It's more efficient, and thus necessary here, to group these together, to give `(node(id:1,2,3))`. One way to do that would be to replace the line you have above (L#809 fof opq.R) with something like:

``` r
if (!is.null (opq$id)) { # opq with opq_osm_id

    id_types <- data.frame (type = opq$id$type, id = opq$id$id)
    id_types <- lapply (split (id_types, f = factor (id_types$type)), function (i) {
        sprintf (" %s(id:%s);\n", i$type [1], paste0 (i$id, collapse = ","))
    })
    id_types <- paste0 (id_types, collapse = "")
        
    res <- paste0 (opq$prefix, id_types, opq$suffix)

}
```

Note that `split` groups things regardless of their original order, so even if you submit `"way", "rel", "way", "rel"`, they'll still just be split into the appropriate two groups.

---

Another minor issue is that the initial checks also need to be improved. You've currently just added `all` to the `length(id) == 1 && grepl(...)` bit, but that's not quite sufficient here. Those checks need to be replaced with checks to ensure that either `length(type) == 1` *or* `length(type) == length(id)`, and then proceed to ensure all `type` values are the 3 allowed.

---

I think those two tweaks should pretty much get us there - let me know when that's done, and i'll aim to merge asap after that. Thanks!

jmaspons:
I think it's ready.

For the initial checks, I removed the `length(id) == 1` as since https://github.com/ropensci/osmdata/pull/282 opq can work with multiple ids, now also for id in type/id format.

Also, instead of testing `length(type) == length(id)`, I test for `length (id) %% length (type) != 0`, a condition for recycling when creating the data.frame in opq_string

R/opq.R
```diff
@@ -794,8 +806,13 @@ opq_string_intern <- function (opq, quiet = TRUE) {
 
     } else if (!is.null (opq$id)) { # opq with opq_osm_id
 
-        id <- paste (opq$id$id, collapse = ",")
-        id <- sprintf (" %s(id:%s);\n", opq$id$type, id)
+        typeId <- data.frame (type=opq$id$type, id=opq$id$id)
```
mpadge, line 809:
```suggestion
        type_id <- data.frame (type=opq$id$type, id=opq$id$id)
```

R/opq.R
```diff
@@ -794,8 +806,13 @@ opq_string_intern <- function (opq, quiet = TRUE) {
...
+        typeId <- split(typeId, typeId$type)
```
mpadge, line 810:
```suggestion
        type_id <- split(type_id, type_id$type)
```

R/opq.R
```diff
@@ -794,8 +806,13 @@ opq_string_intern <- function (opq, quiet = TRUE) {
...
+        id <- mapply (function(type, ids){
+          paste0 (" ", type, "(id:", paste (ids, collapse=","), ");\n")
+        }, type=names(typeId), ids=typeId)
```
mpadge, line 813:
```suggestion
        }, type=names(type_id), ids=type_id)
```

jmaspons:
snake_case done!

Issue #252, opened 2021-11-23
[FEATURE] Add "PRO" features: custom queries


Mashin6:
First of all, awesome job with parsing OSM data and making use of overpass user friendly!

Although most of the use cases are just a simple filter queries, I think it would be useful to have an option to write custom queries in overpass QL and pass it to functions like `osmdata_sf`. Right now I have to do it the hacky way by modifying `$prefix` and `$features` in `overpass_query` object.

Second, would be to add function `osmdata_df()` that would parse the OP output into regular data.frame object. That way queries like this, which return csv format would possible. This would allow afterwards doing `left_join` on data from `osmdata_sf()` query to have super simple way of e.g. plot count statistics per area object.
```
[out:csv(name,counts;false;",")][timeout:3600];
{{geocodeArea:Southeastern Connecticut COG}}->.a;

rel[admin_level=8](area.a);
map_to_area;

foreach -> .town(
  nwr[natural=tree](area.town)->.objects;
  make stat
    name = town.set(t["name"]),
    counts = objects.count(nwr);
  out;
  );
```

mpadge:
Thanks @Mashin6. For you first question, there is the [the `opq_string()` function](https://docs.ropensci.org/osmdata/reference/opq_string.html). As described [in the main vignette](https://docs.ropensci.org/osmdata/articles/osmdata.html#extracting-osm-data-from-a-query), query strings can be passed directly to the various `osmdata_...` functions.

Your second question is interesting, but may be hard to generalise. I'll think about it and report back - feel free to add any thoughts in the meantime about how you think this might be generalised.

Mashin6:
Oh cool. I didn't realize `osmdata_sf()` can take other things then just a query object.

To be honest this is quite advanced use of OP and there are many ways `foreach;` and `for()` can be used. The simplest implementation would be to have `osmdata_df()` function that parses output of `[out:csv]` type queries.
But if you think that "casual" users would benefit from this then I was thinking adding some more functions would generalize my above example.

```
getbb("Southeastern Connecticut COG", featuretype = "boundary") %>% 
    opq() %>%
    add_csv_fields(list("name" = "set(t[\"name\"])", "counts" = "count(nwr)"), header = FALSE) %>%
    add_osm_topfeature(key="admin_level", value="8") %>%
    add_osm_feature(key="natural", value="tree") %>%
    osmdata_df()
```
`add_csv_fields` would take care of the header and mapping of methods applied to data in `make` part.
[...]

Issue #179, opened 2019-07-25
How to include Date Diff Adiff queries? 


neogeomat:
how to query attic date using the date, diif , adiff parameters as stated here?
https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL#date

mpadge:
Hey, I didn't realise that those QL features had been implemented - thanks so much for the info. At the moment, this is not possible, but i'll try to implement it soon.

neogeomat:
I wrote this much, couldn't get the datetime_string inside quotes as required in the query.
```
add_date <- function (opq,datetime=Sys.time())
{
  datetime_string <- as.character(t,format = "%Y-%m-%dT%H:%M:%SZ")
  opq$prefix <- paste('[date:',datetime_string,']',opq$prefix,sep = "")
  print(opq$prefix)
}
add_date(op)
```

mpadge:
You'd be most welcome to submit a PR if you'd like. You can easily insert quotes by escaping: `\"`. Until you or I implement this, a workaround would be to call `opq()` separately, then modify the string in there by adding `"date:.."`, then pass it to `add_osm_feature() %>% osmdata_sc/f/p()`.

TomBor:
Would be great to have this functionality!

mpadge:
Relevant news from [`r-devel` NEWS](https://stat.ethz.ch/R-manual/R-devel/doc/html/NEWS.html):

> R now provides a syntax for specifying raw character constants similar to the one used in C++: r"(...)" with ... any character sequence, except that it must not contain the closing sequence )". This makes it easier to write strings that contain backslashes or both single and double quotes. For more details see ?Quotes.

Robinlovelace:
Impressive!

mpadge:
That commit implements `opq()` parameters `datetime` (for dates) and `datetime2` (for differences in data between 2 dates). Re-opened to document in vignette.

mpadge:
Also have to add the `datetime` values to the output metadata

jmaspons:
Adiff queries still missing. It requires to parse a different xml response format containing the new and old status of the objects. I would reopen the issue

mpadge:
[...]

PR #285, opened 2022-11-30
Implement osmdata_data_frame supporting `out meta;` and `adiff` queries 

jmaspons:
Works and seems that doesn't break anything (check is fine). I will add tests if you like the idea.


mpadge:
See also #252, which should be retained as the "original" vision of this kind of function, courtesy of @Mashin6. No XML filtering; no nodes, ways, rels; just data. So any `osmdata_data_frame()` function should first of all do that, and then subsequently be overloaded for the kinds of queries envisioned here if desired/necessary. That original vision was to provide access to overpass queries which are intended to return tabular data, and tabular data only.

@jmaspons Would you be able to modify this to reflect #252, and return that kind of data.frame, and yet still incorporate the functionality suggested here?

jmaspons:
> See also #252, which should be retained as the "original" vision of this kind of function, courtesy of @Mashin6. No XML filtering; no nodes, ways, rels; just data. So any `osmdata_data_frame()` function should first of all do that, and then subsequently be overloaded for the kinds of queries envisioned here if desired/necessary. That original vision was to provide access to overpass queries which are intended to return tabular data, and tabular data only.

I think the proposal is unrelated as it affects mostly `opq` and how to allow the user to change prefix and suffix, which I am also interested in for other use cases. Let's talk about it there.

> @jmaspons Would you be able to modify this to reflect #252, and return that kind of data.frame, and yet still incorporate the functionality suggested here?

Once #252 is implemented, it should be easy to handle the case by parsing the opq$suffix or whatever flag that the implementation end up with, but I think it's premature while there are no option for different prefixes and suffixes. I will help to modify osmdata_data_frame when needed.

P.S.: reading your post, I interpret that you prefer `osmdata_data_frame()` as a function name. I'll do it

jmaspons:
`osmdata_data_frame` works with `adiff` queries, partially solving #179

mpadge:
@jmaspons I'll get to this one asap, but may take a couple of days.

R/get-osmdata.R
```diff
@@ -256,6 +303,34 @@ get_metadata <- function (obj, doc) {
     return (obj)
 }
 
+is_datetime <- function (x) {
```
mpadge, line 306: This whole function can be written as something like:

``` r
is_datetime <- function (x) {
    ptn <- "[0-9]{4}(\\-|\\/)[0-9]{2}(\\-|\\/)[0-9]{2}\\s*[0-9]{1,2}\\:[0-9]{1,2}\\:[0-9]{1,2}"
    grepl (ptn, x)
}
```
That is maybe a bit of overkill, as the timestamps passed to this are only those generated by overpass, right? In which case they are more standardised that that assumes, so could be simplified to:
``` r
    ptn <- "[0-9]{4}\\-[0-9]{2}\\-[0-9]{2}\\s*[0-9]{2}\\:[0-9]{2}\\:[0-9]{2}"
```

mpadge (changes requested):
Apart from that one comment, this looks great!

R/get-osmdata.R
jmaspons: You are right, the inputs should come from an `opq` object, thus, the datetimes already have passed `check_datetime`. I modified a bit the regex to pass the tests (no need to escape - nor :):

```r
ptn <- "[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}[A-Z]"
```

I don't get what ptn stand for, but it will work.

R/get-osmdata.R
```diff
@@ -305,7 +305,7 @@ get_metadata <- function (obj, doc) {
 
 is_datetime <- function (x) {
 
-    ptn <- "[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}[A-Z]"
+    ptn <- "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}[A-Z]$"
```
mpadge, line 308: Technically that's okay, but leaving in former state will allow the function to be more generally adapted to other cases. All timestamps returned as part of the OSM data set indeed have the form of the latter suggestion, but timestamps are, for example, also returned by the status queries - have a look at `overpass_status()`. The former version is highly specific anyway, and will allow timestamps to be extracted from general strings via regex matches if we need to. So i'd suggest leaving as was, unless you've encountered issues with that form?

jmaspons: This new function just mimicry the behaviour of check_datetime https://github.com/ropensci/osmdata/blob/5c44982863b2e70fa4b99c219125aae457e66524/R/opq.R#L127-L152 (which can be simplified using regexp for a more ecological implementation) and is only used in get-osmdata.R:get_metadata(). If it has to be used in other places to match datetimes in a string, I would rename the function to contain_datetime. As you wish.

mpadge: Yeah, sound good, so we'll leave as is. Thanks!

mpadge:
@jmaspons This seems good to merge now. Did you have any other changes in mind? If not, then I'll merge straight away. I had a play around with it - this really is a very nice addition! I think i'll end up using this quite a bit.

jmaspons:
Great! Go ahead with the merge. I will rebase the other PR on top of this one. Do you have any preference for the order of PR from #292 ? If not, I will go for #288

PR #286, opened 2022-12-05
Server-side filtering by area

jmaspons:
Addresses #256. It allows passing areas to `bbox` parameter in `opq` described by osm type and id (eg. `relation(id:1, 2, 3); way(id:2)`). New functions `getarea` and `area_to_string` help to generate the areas, and they follow the patterns of `getbb` and `bbox_to_sting`.

Despite that the implementation works, there are some naming inconsistencies which require to break the API to fix them, hopefully in a backward compatible way. Bbox are areas, but areas are not always bbox. To fix the inconsistencies, I would:

- [ ] rename `bbox` parameter to `area` in `opq`.
- [ ] restruture `getbb` and `getarea` moving the different `format_out` cases where they belong
    - [ ]  `polygon` and `polygon_sf` moved to `getarea`.
    - [ ]  `data.frame` -> `get_nominatim_query` to the public API
- [ ] rename files getbb.r and test-getbb.r

If agreed about the API break, I can make another PR afterwards.


jmaspons:
https://github.com/ropensci/osmdata/pull/286/commits/c876e9064c267249a40073cfb8a34e640532f637 should be cherry-picked to main if the PR is not merged

jmaspons:
Thanks for the feedback! I don't have strong opinions about the nomenclature, but somehow we have to distinguish the two current forms to crop the data: the bounding box as defined by four coordinates, and the area to crop using an arbitrary area using the overpass function [map_to_area](https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL#Map_way/relation_to_area_(map_to_area)). Some grep could be enough while keeping the different data formats to opq$bbox.

The less invasive way could be to add a new `format_out` mode to `getbb` (eg. `osm_type_id`), remove the `getarea` and `area_to_string` functions and adapt `bbox_to_string` to the new format. Is that fine for you?

mpadge:
> 
> The less invasive way could be to add a new `format_out` mode to `getbb` (eg. `osm_type_id`), remove the `getarea` and `area_to_string` functions and adapt `bbox_to_string` to the new format. Is that fine for you?

That sounds like a fine solution :+1:

jmaspons:
> > The less invasive way could be to add a new `format_out` mode to `getbb` (eg. `osm_type_id`), remove the `getarea` and `area_to_string` functions and adapt `bbox_to_string` to the new format. Is that fine for you?
> 
> That sounds like a fine solution +1

Changes done and rebased

R/opq.R
```diff
@@ -245,13 +264,11 @@ add_osm_feature <- function (opq,
         stop ("key must be provided")
     }
 
-    if (is.null (bbox) & is.null (opq$bbox)) {
+    if (is.null (bbox) && is.null (opq$bbox)) {
```
mpadge, line 267: Thanks for catching those!

mpadge (changes requested):
That looks great @jmaspons! Can you please just change opq.RL#857 from `if (!quiet)` to `if (!quiet && !map_to_area)`, to suppress that message for `map_to_area` queries?

jmaspons:
> Can you please just change opq.RL#857 from `if (!quiet)` to `if (!quiet && !map_to_area)`, to suppress that message for `map_to_area` queries?
 
Sure, but the bbox can be as big and the query as heavy as with a squared bounding box. Still?

R/opq.R
jmaspons: You are welcome. I spend some time reading the code :)

mpadge:
> > Can you please just change opq.RL#857 from `if (!quiet)` to `if (!quiet && !map_to_area)`, to suppress that message for `map_to_area` queries?
> 
> Sure, but the bbox can be as big and the query as heavy as with a squared bounding box. Still?

Yeah, good point. I guess then leave as is, if you think that's advisable.

mpadge:
Then i think just one more requst @jmaspons: Could you please add yourself as "ctb" in [the 'DESCRIPTION' file](https://github.com/ropensci/osmdata/blob/main/DESCRIPTION)? Thanks

jmaspons:
Done! Thanks :)

jmaspons:
I hope so. I will use the package, so I'm interested in it working well

mpadge:
Thank you for the great contribution!

PR #289, opened 2022-12-07
Fix the logic of parameter checks for osmdata_*(doc = doc)

jmaspons:
Now the code match the documentation and works for calls with `doc` and no `q` parameter.
TODO: `osmdata_sc` fails because it requires the bbox

### DIFF
--- a/R/get-osmdata.R
+++ b/R/get-osmdata.R
@@ -86,8 +86,8 @@ osmdata_xml <- function (q, filename, quiet = TRUE, encoding) {
         encoding <- "UTF-8"
     }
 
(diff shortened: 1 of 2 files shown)


mpadge:
Thanks for that - another great contribution! I'm happy to merge now, if that's okay with you? If so, then could you also open a new issue about the `osmdata_sc` issue, to ensure we fix that too. Thanks :+1:

jmaspons:
Yes, I think it's ready once the checks pass with the new tests

jmaspons:
Rebased and ready for review @mpadge

PR #294, opened 2022-12-15
Optimize osmdata data frame

jmaspons:
Preallocate tag columns in a matrix and use xml2 to extract the values for all the nodes in one call for each column.

``` r
library(xml2)
library(osmdata)
# devtools::load_all (".")
q0 <- opq (bbox = c(-0.27, 51.47, -0.20, 51.50))
q1 <- add_osm_feature (q0, key = "name", value = "River Thames")

doc <- osmdata_xml (q1, filename = test_path ("export.osm"))
doc <- xml2::read_xml ( test_path ("export.osm"))
stringsAsFactors <- FALSE
osm_obj <- xml2::xml_find_all (doc, ".//node|.//way|.//relation")

mb <- list()

## Implementations of the critical code ----
impl <- list()
impl$baseline <- function(osm_obj=osm_obj) {
    tags <- xml2::xml_find_all (osm_obj, xpath = ".//tag", flatten = FALSE)
    tagsL <- lapply (tags, function(x) {
        tag <- xml2::xml_attrs (x)
        tag <- structure (lapply (tag, function(y) y[["v"]]),
                          names = vapply(tag, function(y) y["k"], character(1)))
        list2DF (tag, nrow = 1)
    })
    df <- do.call(osmdata:::rbind_add_columns, c (tagsL, list (stringsAsFactors = stringsAsFactors)))
    return (df[, sort(names(df))])
}

impl$preallocate <- function(osm_obj=osm_obj) {
    tags <- xml2::xml_find_all (osm_obj, xpath = ".//tag", flatten = FALSE)
    tagsU <- xml2::xml_find_all (osm_obj, xpath = ".//tag")
    col_names <- sort(unique(xml2::xml_attr(tagsU, attr="k")))
    m <- matrix(nrow = length (tags), ncol = length (col_names), dimnames = list(NULL, col_names))
    for (i in seq_along(tags)) {
        tag <- xml2::xml_attrs (tags[[i]])
        cols <- vapply (tag, function(x) x["k"], FUN.VALUE = character(1))
        m[i, cols] <- vapply (tag, function(x) x["v"], FUN.VALUE = character(1))
    }
    return (as.data.frame(m, stringsAsFactors = stringsAsFactors))
}

impl$preallocate_sel <- function(osm_obj=osm_obj) {
    tags <- xml2::xml_find_all (osm_obj, xpath = ".//tag", flatten = FALSE)
    tagsU <- xml2::xml_find_all (osm_obj, xpath = ".//tag")
    col_names <- sort(unique(xml2::xml_attr(tagsU, attr="k")))
    m <- matrix(nrow = length (tags), ncol = length (col_names), dimnames = list(NULL, col_names))
    has_tags <- which (vapply(tags, length, FUN.VALUE = integer(1)) > 0)
    for (i in has_tags) {
        tag <- xml2::xml_attrs (tags[[i]])
        tagV <- vapply (tag, function(x) x, FUN.VALUE = character(2))
        m[i, tagV[1, ]] <- tagV[2, ]
    }
    return (as.data.frame(m, stringsAsFactors = stringsAsFactors))
}

impl <- lapply(impl, compiler::cmpfun)


## Benchmark ----
mb <- microbenchmark::microbenchmark(impl$baseline(osm_obj=osm_obj),
                                     impl$preallocate(osm_obj=osm_obj),
                                     impl$preallocate_sel(osm_obj=osm_obj),
                                     times = 20, check = "equal")
ggplot2::autoplot(mb)
```

``` r
mb
#> Unit: milliseconds
#>                                     expr       min        lq      mean
#>         impl$baseline(osm_obj = osm_obj) 6260.8424 6382.3150 6776.0929
#>      impl$preallocate(osm_obj = osm_obj)  367.2648  391.6788  434.5904
#>  impl$preallocate_sel(osm_obj = osm_obj)  255.9076  277.8087  308.8099
#>     median        uq       max neval cld
#>  6576.6413 6980.1697 8472.3096    20   b
#>   414.1598  462.8838  642.2182    20  a 
#>   299.4965  321.3393  458.2554    20  a
```

The new implementation is based on the `preallocate_sel` code. The benchmark shows that is up to 20 times faster than the current implementation.

### DIFF
--- a/R/get-osmdata.R
+++ b/R/get-osmdata.R
@@ -758,164 +758,67 @@ xml_adiff_to_df <- function (doc,
         ))
     }
 
-    action_type <- xml2::xml_attr (osm_actions, attr = "type")
+    osm_obj <- xml2::xml_find_all (osm_actions, ".//node|.//way|.//relation")
 
-    df_l <- mapply (function (action, type) {
-        osm_obj <- xml2::xml_find_all (action, ".//node|.//way|.//relation")
-        osm_type <- xml2::xml_name (osm_obj)
-        osm_id <- xml2::xml_attr (osm_obj, attr = "id")
+    tagsU <- xml2::xml_find_all (osm_actions, xpath = ".//tag")
+    col_names <- sort (unique (xml2::xml_attr (tagsU, attr="k")))
+    m <- matrix (nrow = length (osm_obj), ncol = length (col_names),
+                 dimnames = list (NULL, col_names))
 
-        if (type == "modify") {
+    tags <- xml2::xml_find_all (osm_obj, xpath = ".//tag", flatten = FALSE)
+    has_tags <- which (vapply (tags, length, FUN.VALUE = integer (1)) > 0)
+    for (i in has_tags) {
+        tag <- xml2::xml_attrs (tags[[i]])
+        tagV <- vapply (tag, function (x) x, FUN.VALUE = character (2))
+        m[i, tagV[1, ]] <- tagV[2, ]
+    }
 
-            dates <- c (datetime_from, datetime_to)
-            tags <- xml2::xml_find_all (
-                osm_obj,
-                xpath = ".//tag",
-                flatten = FALSE
-            )
-            tags_l <- mapply (function (x, adiff_date) {
-                tag <- xml2::xml_attrs (x)
-                tag <- structure (lapply (tag, function (y) y [["v"]]),
-                    names = vapply (tag, function (y) y ["k"], character (1))
-                )
-                list2DF (c (list (
-                    adiff_action = "modify",
-                    adiff_date = adiff_date
-                ), tag), nrow = 1)
-            }, x = tags, adiff_date = dates, SIMPLIFY = FALSE)
-            df <- do.call (
-                rbind_add_columns,
-                c (tags_l, list (stringsAsFactors = stringsAsFactors))
-            )
+    osm_type <- xml2::xml_name(osm_obj)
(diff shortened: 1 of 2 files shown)


mpadge:
That looks awesome. One Q for you @jmaspons: do you speak C++? This package has an internally bundled version of ["rapidxml"](https://rapidxml.sourceforge.net/), which is way faster than `xml2`. It's exposed here via [the `XmlData` class](https://github.com/ropensci/osmdata/blob/9b8f306d62b6bef49e7aed3df0d3aae18bb6113a/src/osmdata.h#L129-L175), which means you just need to call [the `parseXML` function](https://github.com/ropensci/osmdata/blob/main/src/common.cpp) and you get a pointer to all of the internal data. I suspect that should be way faster again, and should at least be tried in these benchmarks. No problems if you're not able - just let me know, and I'll add some code to this PR to include that too.

jmaspons:
I know the basics, but I'm not fluent in C++. I had a look to the code and get lost. With some guidance (as you just did), I can try and take the opportunity to practice. Anyway, if you have time and want to implement it, go for it.

jmaspons:
I've tried (https://github.com/jmaspons/osmdata/tree/osmdata_data_frame-cpp) but I can't get proper results. It seems that the results still contain spatial data. The main problem I faced is how to debug the code. Following https://github.com/wch/r-debug/blob/master/debugging-r.md didn't work, and I failed to generate a proper cmake or make file to work with a c++ IDE.

I would merge the patch as is, as it is already a big improvement, and leave further optimizations for later.

PR #298, opened 2022-12-30
add C++ implementation of osmdata_data_frame

mpadge:
@jmaspons Here is a C++ implemtation of `osmdata_data_frame`. The following code includes the final R function which can replace your current code for `xml_to_df()` if you think this is okay, and if you approve this PR.

``` r
library(xml2)
# devtools::load_all (export_all = TRUE)
library (osmdata)
#> ℹ Loading osmdata
q0 <- opq (bbox = c(-0.27, 51.47, -0.20, 51.50))
q1 <- add_osm_feature (q0, key = "name", value = "River Thames")

f <- test_path ("export.osm")
if (!file.exists (f)) {
    doc <- osmdata_xml (q1, filename = f)
}
doc <- xml2::read_xml (f)

impl <- list ()
impl$preallocate_sel <- function (doc) {
    xml_to_df (doc)
}
impl$get_osmdata_df <- function (doc) {
    xml_to_df_cpp (doc)
}
impl <- lapply(impl, compiler::cmpfun)

mb <- microbenchmark::microbenchmark (
    impl$preallocate_sel (doc),
    impl$get_osmdata_df (doc),
    times = 20L,
    check = NULL # Not equal, because 'preallocate_sel' does not return 'osm_id' column
)
ggplot2::autoplot (mb)
#> Coordinate system already present. Adding new coordinate system, which will
#> replace the existing one.
```

<sup>Created on 2022-12-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


mpadge:
@jmaspons A couple of notes:

1. I'll request that you review this code once you've commented in here, and had a go yourself to see if you find any issues.
2. The main difference to your `xml_to_df()` function is that yours has the explicit code for timestamps, changesets, and the other stuff related to (a)diff queries. I haven't actually run the new version, but the C++ code should simply pick all of those values up anytime they are present in the `doc`, so (I'm pretty sure ...) this version does not need that code.

Let me know what you think. Thanks :+1:

jmaspons:
Cool, less carbon to the atmosphere  :) @mpadge ! I'm reviewing the code and will response asap, but it's probably ok as all tests pass 

The behaviour you describe in your 2nd point about the timestamps, changesets, & Co. , is true for all other osmdata_* implemented in c++? If not, would be a good improvement. If yes, we can fix the `opq` docs that state that `out="meta"` only works for `osmdata_xml` and `osmdata_data_frame`, and perhaps add some tests.

jmaspons (changes requested):
Some comments on the code. I have not tested and it seems that the new code is not really called yet. Let's change xml_to_df for xml_to_df_cpp and see what happens with the tests.

I think that the code is not ready for `adiff` queries, isn't it?

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+        rel_id_mp.push_back (std::to_string (itr->id));
+
+        osm_convert::get_value_mat_rel (itr, unique_vals, kv_mat, count++);
+    }
+
+    Rcpp::DataFrame kv_df;
+    if (rel_id_mp.size () > 0)
+    {
+        kv_mat.attr ("names") = unique_vals.k_rel;
+        kv_mat.attr ("dimnames") = Rcpp::List::create (rel_id_mp, unique_vals.k_rel);
```
jmaspons, line 96:
```suggestion
```
row names are not relevant a can be removed

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+        {
+            const std::string &key = kv_iter->first;
+            unsigned int ndi = unique_vals.k_point_index.at (key);
+            kv_mat (count, ndi) = kv_iter->second;
+        }
+        count++;
+    }
+    if (unique_vals.k_point.size () > 0)
+    {
+        kv_mat.attr ("dimnames") = Rcpp::List::create (ptnames, unique_vals.k_point);
```
jmaspons, line 183:
```suggestion
        kv_mat.attr ("names") = unique_vals.k_point;
```
row names are not relevant a can be removed

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+        auto wj = ways.find (*wi);
+        osm_convert::get_value_mat_way (wj, unique_vals, kv_mat, count);
+        count++;
+    }
+
+    kv_df = R_NilValue;
+    if (way_ids.size () > 0)
+    {
+        kv_mat.attr ("names") = unique_vals.k_way;
+        kv_mat.attr ("dimnames") = Rcpp::List::create (waynames, unique_vals.k_way);
```
jmaspons, line 140:
```suggestion
```
row names are not relevant a can be removed

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+    std::vector <std::string> waynames;
+    waynames.reserve (way_ids.size ());
+
+    Rcpp::CharacterMatrix kv_mat (Rcpp::Dimension (nrow, ncol));
+    std::fill (kv_mat.begin (), kv_mat.end (), NA_STRING);
+    unsigned int count = 0;
+    for (auto wi = way_ids.begin (); wi != way_ids.end (); ++wi)
+    {
+        Rcpp::checkUserInterrupt ();
+        waynames.push_back (std::to_string (*wi));
```
jmaspons, line 130:
```suggestion
```

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+//' 
+//' @noRd 
+void osm_df::get_osm_ways (Rcpp::DataFrame &kv_df,
+        const std::set <osmid_t> &way_ids, const Ways &ways,
+        const UniqueVals &unique_vals)
+{
+
+    size_t nrow = way_ids.size (), ncol = unique_vals.k_way.size ();
+    std::vector <std::string> waynames;
+    waynames.reserve (way_ids.size ());
```
jmaspons, line 122:
```suggestion
```

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+//'
+//' @return A dual Rcpp::List, the first of which contains the multipolygon
+//'         relations; the second the multilinestring relations.
+//' 
+//' @noRd 
+Rcpp::DataFrame osm_df::get_osm_relations (const Relations &rels, 
+        const UniqueVals &unique_vals)
+{
+    std::vector <std::string> ids_mp, rel_id_mp; 
+    std::vector <std::string> roles;
```
jmaspons, line 73:
```suggestion
```
not used

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+//'       unique IDs and keys for each kind of OSM object (nodes, ways, rels).
+//'
+//' @return A dual Rcpp::List, the first of which contains the multipolygon
+//'         relations; the second the multilinestring relations.
+//' 
+//' @noRd 
+Rcpp::DataFrame osm_df::get_osm_relations (const Relations &rels, 
+        const UniqueVals &unique_vals)
+{
+    std::vector <std::string> ids_mp, rel_id_mp;
```
jmaspons, line 72:
```suggestion
    std::vector <std::string> ids_mp; 
```
Skip rel_id_mp var. As I understand, is only used to build the rownames and be substituted in the if statement

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+Rcpp::DataFrame osm_df::get_osm_relations (const Relations &rels, 
+        const UniqueVals &unique_vals)
+{
+    std::vector <std::string> ids_mp, rel_id_mp; 
+    std::vector <std::string> roles;
+
+    const unsigned int nmp = rels.size ();
+
+    size_t ncol = unique_vals.k_rel.size ();
+    rel_id_mp.reserve (nmp);
```
jmaspons, line 78:
```suggestion
```
Skip rel_id_mp var

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+    Rcpp::CharacterMatrix kv_mat (Rcpp::Dimension (nmp, ncol));
+    unsigned int count = 0;
+
+    for (auto itr = rels.begin (); itr != rels.end (); ++itr)
+    {
+        Rcpp::checkUserInterrupt ();
+
+        rel_id_mp.push_back (std::to_string (itr->id));
```
jmaspons, line 87:
```suggestion
```
Skip rel_id_mp var

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+        osm_convert::get_value_mat_rel (itr, unique_vals, kv_mat, count++);
+    }
+
+    Rcpp::DataFrame kv_df;
+    if (rel_id_mp.size () > 0)
```
jmaspons, line 93:
```suggestion
    if (nmp > 0)
```
Skip rel_id_mp var

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+void osm_df::get_osm_nodes (Rcpp::DataFrame &kv_df,
+        const Nodes &nodes, const UniqueVals &unique_vals)
+{
+    size_t nrow = nodes.size (), ncol = unique_vals.k_point.size ();
+
+    Rcpp::CharacterMatrix kv_mat (Rcpp::Dimension (nrow, ncol));
+    std::fill (kv_mat.begin (), kv_mat.end (), NA_STRING);
+
+    std::vector <std::string> ptnames;
+    ptnames.reserve (nodes.size ());
```
jmaspons, line 164:
```suggestion
```
skip ptnames var

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+    unsigned int count = 0;
+    for (auto ni = nodes.begin (); ni != nodes.end (); ++ni)
+    {
+        if (count % 1000 == 0)
+            Rcpp::checkUserInterrupt ();
+
+        ptnames.push_back (std::to_string (ni->first));
```
jmaspons, line 171:
```suggestion
```
skip ptnames

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+        count++;
+    }
+    if (unique_vals.k_point.size () > 0)
+    {
+        kv_mat.attr ("dimnames") = Rcpp::List::create (ptnames, unique_vals.k_point);
+        kv_df = osm_convert::restructure_kv_mat (kv_mat, false);
+    } else
+        kv_df = R_NilValue;
+
+    ptnames.clear ();
```
jmaspons, line 188:
```suggestion
```
skip ptnames

jmaspons (commented):
Remove row names attributes and associated variables

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+     * 2. Extract OSM ways
+     * --------------------------------------------------------------*/
+
+    std::set <osmid_t> way_ids;
+    for (auto itw = ways.begin (); itw != ways.end (); ++itw)
+    {
+        way_ids.insert ((*itw).first);
+    }
+
+    Rcpp::List polyList (way_ids.size ());
```
jmaspons, line 235:
```suggestion
```
not used

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,259 @@
...
+//' Store key-val pairs for OSM ways as a list/data.frame
+//'
+//' @param kv_df Pointer to Rcpp::DataFrame to hold key-value pairs
+//' @param way_ids Vector of <osmid_t> IDs of ways to trace
+//' @param ways Pointer to all ways in data set
+//' @param nodes Pointer to all nodes in data set
+//' @param unique_vals pointer to all unique values (OSM IDs and keys) in data set
+//' 
+//' @noRd 
+void osm_df::get_osm_ways (Rcpp::DataFrame &kv_df,
```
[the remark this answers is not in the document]
mpadge: Yeah, i intend to do that, but just wanted to get it working first. Thanks

src/osmdata-data_frame.cpp
mpadge: There's no real disadvantage to keeping it, and since R will insert dimnames as soon as the object is returned to the R workspace, I think it's better to keep that (and the following) lines, so at least the dimnames hold useful information.

- *Disadvantage* Only a likely entirely negligible overhead in constructing the list
- *Advantage* Having dimnames that hold useful  information, even if that is entirely redundant.

mpadge:
@jmaspons Thanks so much for your review of the code. Note that the above commit restores all of the C++ `dimnames` construction lines. I had already accepted your suggestions to remove all of those lines, but forgot that the `restructre_kv_mat` function, which is called at the end of all functions here, actually needs and uses that information: https://github.com/ropensci/osmdata/blob/e9dedbb79d0995f259caa3ea3e79648f19515158/src/convert-osm-rcpp.cpp#L149-L150
So all has now been reinstated.

mpadge:
@jmaspons And the above commit acually reverts back to your suggestion for all C++ internal functions to directly return `Rcpp::DataFrame` objects. There was a reason why those were all passed by reference in the other (sf/sc) versions, but I can't remember what that was, and in this case, it's definitely cleaner to do it as you suggested, so thanks.

src/osmdata-data_frame.cpp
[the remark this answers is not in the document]
mpadge: Resolved as described below. Thanks.

R/get-osmdata.R
```diff
@@ -745,6 +746,46 @@ xml_to_df <- function (doc, stringsAsFactors = FALSE) {
...
+    df <- do.call (rbind, df)
+
+    if (nrow (df) == 0) {
+        df <- data.frame (
+            osm_type = character (),
+            osm_id = character (),
+            stringsAsFactors = stringsAsFactors
+        )
+    }
+
```
jmaspons, line 784:
```suggestion

    rownames (df) <- NULL

```
Remove uninformative row names

jmaspons:
Good! I'm testing the correctness of the implementation. For the basic case without metadata nor adiff queries it produces the same results as a version in main except for the column order (suggested but more work in the c++ side required), row names (suggested above) and row order (it's ok as is).

For queries with metadata, it should be extracted from the attributes of each object, as for the osm_id (eg. `<way id="1056516701" version="1" timestamp="2022-05-01T21:51:18Z" changeset="120436630" uid="508" user="Welshie">`). Then I rename attributes prepending `osm_` to the metadata columns. Perhaps is a good idea to prepend also a _ to reduce the chances of clashes with key names, at least in the c++ party (it already happens with [osm_id](https://taginfo.openstreetmap.org/keys/osm_id))

Adiff queries also require more work, but perhaps we can skip this for now, as the xml structure is quite different from regular queries.

R/get-osmdata.R
```diff
@@ -745,6 +746,46 @@ xml_to_df <- function (doc, stringsAsFactors = FALSE) {
...
+    }
+    if (nrow (res$ways_kv) > 0L) {
+        res$ways_kv$osm_type <- "way"
+    }
+    if (nrow (res$rels_kv) > 0L) {
+        res$rels_kv$osm_type <- "relation"
+    }
+
+    nms <- sort (unique (unlist (lapply (res, names))))
+    nms1 <- c ("osm_type", "osm_id", "name")
```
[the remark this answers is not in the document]
jmaspons: Oh, the suggested change is wrong. The intersection should have `nms` as a second parameter because the resulting order of the result follows the first parameter.

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+        rel_ids.push_back (std::to_string (itr->id));
+
+        osm_convert::get_value_mat_rel (itr, unique_vals, kv_mat, count++);
+    }
+
+    Rcpp::DataFrame kv_df;
+    if (nmp > 0)
+    {
+        kv_mat.attr ("names") = unique_vals.k_rel;
```
jmaspons, line 92:
```suggestion
```
`names` attribute is the same as the second dimension of the `dimnames` assigned in the following line

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+        auto wj = ways.find (*wi);
+        osm_convert::get_value_mat_way (wj, unique_vals, kv_mat, count);
+        count++;
+    }
+
+    Rcpp::DataFrame kv_df = R_NilValue;
+    if (way_ids.size () > 0)
+    {
+        kv_mat.attr ("names") = unique_vals.k_way;
```
jmaspons, line 140:
```suggestion
```
`names` attribute is the same as the second dimension of the `dimnames` assigned in the following line

src/osmdata-data_frame.cpp
```diff
@@ -0,0 +1,267 @@
...
+            kv_mat (count, ndi) = kv_iter->second;
+        }
+        count++;
+    }
+
+    Rcpp::DataFrame kv_df = R_NilValue;
+    if (unique_vals.k_point.size () > 0)
+    {
+        kv_mat.attr ("dimnames") = Rcpp::List::create (ptnames, unique_vals.k_point);
+        kv_mat.attr ("names") = unique_vals.k_point;
```
jmaspons, line 193:
```suggestion
```
`names` attribute is the same as the second dimension of the `dimnames` assigned in the later line

jmaspons (changes requested):
Avoid redundant assignment of the attribute `names` , already set as the second parameter of dimnames

mpadge:
@jmaspons Finally got some time to work on this again. The metadata can be extracted in C++ for no extra overhead, so i want to do all now, then same for adiff data, then i'll merge this. The above commit starts by extracting metadata for nodes, providing a template for the rest.

mpadge:
@jmaspons The above commits get us this far:

``` r
devtools::load_all (export_all = TRUE)
#> ℹ Loading osmdata

impl <- list ()
impl$preallocate_sel <- function (doc) {
    xml_to_df (doc)
}
impl$get_osmdata_df <- function (doc) {
    xml_to_df_cpp (doc)
}
impl <- lapply(impl, compiler::cmpfun)

### No metadata -----

q0 <- opq (bbox = c(-0.27, 51.47, -0.20, 51.50))
q1 <- add_osm_feature (q0, key = "name", value = "River Thames")

f <- test_path ("export.osm")
if (!file.exists (f)) {
    doc <- osmdata_xml (q1, filename = f)
}
doc <- xml2::read_xml (f)

resR <- impl$preallocate_sel (doc)
resCpp <- impl$get_osmdata_df (doc)

waldo::compare (resR, resCpp)
#> ✔ No differences

### With  metadata -----

qmeta <- opq (bbox = c(-0.27, 51.47, -0.20, 51.50), out = "meta")
qmeta <- add_osm_feature (qmeta, key = "name", value = "River Thames")

f <- test_path ("export_meta.osm")
if (!file.exists (f)) {
    doc_meta <- osmdata_xml (qmeta, filename = f)
}
doc_meta <- xml2::read_xml (f)

resR <- impl$preallocate_sel (doc_meta)
resCpp <- impl$get_osmdata_df (doc_meta)

waldo::compare (resR, resCpp)
#> ✔ No differences
```

<sup>Created on 2023-01-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

## TODO:

- [ ] Same for adiff data

PR #299, opened 2023-01-03
Allow to chain  add_osm_feature/s function calls

jmaspons:
Make add_osm_feature(s) do the proper AND and OR additions to the osm feature filters

Before the patch:
* add_osm_feature collapse features from add_osm_features if called first
* if add_osm_feature is called first, features get eliminated after calling add_osm_features

After:
* add_osm_feature add features to each item from opq$features
* add_osm_features add features to the opq$features vector with new items

Bonus: features parameter now is parsed using paste_features which allow
to pass more than one value for each list item


jmaspons:
Ready for review @mpadge

jmaspons:
All green now, @mpadge

mpadge:
Great, thanks @jmaspons :grinning:

PR #300, opened 2023-01-15
Checks in osmdata_* to intercept not implemented query types

jmaspons:
stop cases that crash and warnings for incomplete results.


mpadge:
Thanks, that's a great addition. I'll just  wait until I've merged #298, then merge that over here before finally merging this one again.

jmaspons:
Ok, I see that you are doing good progress there. Perhaps we can remove many of the errors and warnings from this PR, which is great!

jmaspons:
Testing the `osmdata_*` functions again, I see that the changes for implementing meta queries in C++ for `osmdata_data_frame` doesn't affect the other function. Thus, I think it's OK to merge this PR until addressed

mpadge (changes requested):
Thanks @jmaspons, that looks great. Can you please just make one change, and separate that code which is repeated three times into a separate function? something like `check_not_implemented_query_types(q)`, or a shorter equivalent. `check_query_out_types(q)`? Whatever you think is best. Thanks :+1:

jmaspons:
> separate that code which is repeated three times into a separate function

Done! @mpadge

mpadge (approved):
Great, thanks so much.

PR #302, opened 2023-01-26
Update docs and dependencies

jmaspons:
Added missing `osmdata_*` functions, `silicate` as a suggested dependency and `sp` moved from imported to suggested


DESCRIPTION
```diff
@@ -51,6 +50,8 @@ Suggests:
     raster,
     rmarkdown,
     sf,
+    silicate,
```
mpadge, line 53:
```suggestion
```
We don't actually use an silicate functions at all, so that package itself is not needed. The `osmdata_sc` function merely produces output in a format compatible with silicate functionality. Once that's removed, I'll happily approve and merge.

mpadge: And FYI that funciton was developed purely to provide the best input for [the `dodgr` package](https://github.com/atfutures/dodgr).

mpadge (changes requested):
Actually, moving `sp` to Suggests - which is a really great idea!! - then needs each function which actually uses it to first have:
``` r
requireNamespace ("sp")
```
Those 2 functions are
https://github.com/ropensci/osmdata/blob/85cd6954d1e007b0a910195ac941124546fe3898/R/elevation.R#L16
and 
https://github.com/ropensci/osmdata/blob/85cd6954d1e007b0a910195ac941124546fe3898/R/trim-osmdata.R#L63
The latter has an example of how it'd done here - initial lines at the very start of the function. Thanks

mpadge:
@jmaspons Once you address the above, i'll happily approve and merge. I also wanted to arrange a video chat with you, mostly so that i can express my gratitude for all of your work. I don't have contact details for you, so can you please either email me, or you should be able to book a meeting at cal.com / markpadgham. Thanks.

jmaspons:
Hopfully solved! I'll contact you, @mpadge . Let's find some day for next week

R/elevation.R
```diff
@@ -85,6 +88,10 @@ check_elev_file <- function (elev_file) {
 
 check_bbox <- function (dat, r) {
 
+    if (!requireNamespace ("sp", quietly = TRUE)) {
+        stop ('osm_elevation requires sp package:\n\tinstall.packages("sp")')
```
mpadge, line 92: Here too, it's sufficient just to have a simple one-liner, `requireNamespace("sp")`.

R/trim-osmdata.R
```diff
@@ -155,6 +155,10 @@ trim_to_poly_pts <- function (dat, bb_poly, exclude = TRUE) {
 
     if (is (dat$osm_points, "sf")) {
 
+        if (!requireNamespace ("sp", quietly = TRUE)) {
+            stop ('trim_osmdata requires sp package:\n\tinstall.packages("sp")')
```
mpadge, line 159: As in `elevation.R`, better to just have a simple call to `requireNamespace("sp")`.

jmaspons: Done

R/elevation.R
jmaspons: done

R/elevation.R
```diff
@@ -15,13 +15,16 @@
 #' @export
 osm_elevation <- function (dat, elev_file) {
 
-    if (!"raster" %in% loadedNamespaces ()) {
-        requireNamespace ("raster")
-        message (
-            "Elevation data from Consortium for Spatial Information; ",
-            "see http://srtm.csi.cgiar.org/srtmdata/"
+    if (!requireNamespace ("raster", quietly = TRUE)) {
```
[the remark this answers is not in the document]
jmaspons: done

jmaspons:
```
==> Rcpp::compileAttributes()

* Updated R/RcppExports.R
Warning message:
The following packages are referenced using Rcpp::depends attributes however are not listed in the Depends, Imports or LinkingTo fields of the package DESCRIPTION file: sp 
```

Perhaps it requires some more work to move `sp` to suggested packages

mpadge:
Sorry, that's my fault. Just need to delete this line:
https://github.com/ropensci/osmdata/blob/85cd6954d1e007b0a910195ac941124546fe3898/src/osmdata-sp.cpp#L430
That was just there for some historical reason I can no longer remember, but is no longer needed.

mpadge:
Thanks @jmaspons. One more thing that generally needs to be done is to update the NEWS file. I'll merge for now and do that straight away, but we should both try to remember to always update NEWS with each PR :grinning:

PR #303, opened 2023-01-27
Fix key modifications for non-valid column names in osmdata_* functions

### LINKED ISSUE #301: [BUG] `osmdata_*` functions return features with changed tag names for tags with non-valid names
As per [make.names](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/make.names), valid syntactic names only include letters, numbers and the dot or underline characters and starts with a letter. For exemple, many keys in OSM contain also `:` which get transformed to `.`, then `name:ca` becomes `name.ca`

Non-valid names are not a big problem in R. Columns can be accessed anyway by ```df$`key:xx` ``` or `df[, "key:xx"]`. Name changes can be aboided with `data.frame(..., check.names=FALSE)`. I will try to make a PR if you agree with the behavior change.

jmaspons:
FIX #301 and handle cases with clashes between OSM keys and id or metadata columns

### DIFF
--- a/R/get-osmdata.R
+++ b/R/get-osmdata.R
@@ -846,7 +846,7 @@ xml_to_df <- function (doc, stringsAsFactors = FALSE) {
             "renamed by appending `.n`:\n\t",
             paste (setdiff (cols_no_dup, names (df)), collapse = ", ")
         )
-        names (df) <- fix_duplicated_columns (names (df))
+        names (df) <- cols_no_dup
     }
 
     if (nrow (df) == 0) {
--- a/R/get-osmdata.R
+++ b/R/get-osmdata.R
@@ -102,6 +102,39 @@ osmdata_xml <- function (q, filename, quiet = TRUE, encoding) {
     invisible (doc)
 }
 
+fix_duplicated_columns <- function (x) {
+    dup <- duplicated (x)
+    i <- 1
+    while (any (dup)) {
+        x[dup] <- paste0 (x[dup], ".", i)
+        i <- i + 1
+        dup <- duplicated (x)
+    }
+
+    return (x)
+}
+
+fix_columns_list <- function (l) {
+    cols <- lapply (l, names)
+    cols_no_dup <- lapply (cols, fix_duplicated_columns)
+    if (!identical (cols, cols_no_dup)) {
+        warning (
+            "Feature keys clash with id or metadata columns and will be ",
+            "renamed by appending `.n`:\n\t",
+            paste (
+                unique (setdiff (unlist (cols_no_dup), unlist(cols))),
+                collapse = ", "
+            )
+        )
+        l <- mapply (function (x, col) {
+            suppressWarnings (names (x) <- col)
+            x
+        }, x = l, col = cols_no_dup, SIMPLIFY = FALSE)
+    }
+
+    return (l)
+}
+
 #' Return an OSM Overpass query as an \link{osmdata} object in \pkg{sp}
 #' format.
 #'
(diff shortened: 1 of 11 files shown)


jmaspons:
Ready for review, @mpadge ! I'm not sure how silicate objects should handle clashes between tags and id or metadata keys

mpadge:
> I'm not sure how silicate objects should handle clashes between tags and id or metadata keys

That's a good question, but one that i'm happy to worry about later. As said, I'm pretty confident the only use of `osmdata_sc()` is by people interfacing that with `dodgr`, in which case metadata are not used anyway. I'll just keep that point in mind, and maybe address in the future if any issues arise.

R/get-osmdata.R
```diff
@@ -750,54 +792,63 @@ xml_to_df <- function (doc, stringsAsFactors = FALSE) {
...
-    nms1 <- intersect (nms1, nms)
-    nms <- c (nms1, setdiff (nms, nms1))
+    keysL <- lapply ( c ("points_kv", "ways_kv", "rels_kv"), function (x) {
+        out <- names (res[[x]])
+        if (isTRUE (out[1] == "osm_id")) {
+            out <- out[-1] # remove osm_id. Not always present
+        }
+        out
+    })
+    keys<- sort (unique (unlist(keysL)))
```
mpadge, line 802:
```suggestion
    keys <- sort (unique (unlist(keysL)))
```

mpadge:
A few minor suggestions there. Let me know when they're done, and i'll happily merge. This is, as with all the others, a really good PR. Thank you so much! The `fix_duplicated_columns` function is something I hadn't even thought of, but is so obviously important, and excellently impelemted. Great stuff.

R/get-osmdata.R
```diff
@@ -750,54 +792,63 @@ xml_to_df <- function (doc, stringsAsFactors = FALSE) {
...
     rownames (df) <- NULL
 
+    cols_no_dup <- fix_duplicated_columns (names (df))
+    if (!identical (names (df), cols_no_dup)) {
+        warning (
+            "Feature keys clash with id or metadata columns and will be ",
+            "renamed by appending `.n`:\n\t",
+            paste (setdiff (cols_no_dup, names (df)), collapse = ", ")
+        )
+        names (df) <- fix_duplicated_columns (names (df))
```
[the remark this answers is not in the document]
jmaspons: Good catch!

jmaspons:
Thanks for the review, @mpadge ! I think it's ready now

mpadge:
> Thanks for the review, @mpadge ! I think it's ready now

Not quite. The tests are failing because of this:

``` r
devtools::load_all (".", export_all = TRUE)
#> ℹ Loading osmdata
qry <- opq (bbox = c (-0.116, 51.516, -0.115, 51.517))
qry <- add_osm_feature (qry, key = "highway")
res <- osmdata_data_frame (qry)
#> Error in matrix(nrow = nrow(i), ncol = length(keys), dimnames = list(NULL, : non-numeric matrix extent
```

I'll just try to find out what's going wrong.
<sup>Created on 2023-01-31 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/get-osmdata.R
```diff
@@ -750,54 +792,63 @@ xml_to_df <- function (doc, stringsAsFactors = FALSE) {
...
+        if (isTRUE (out[1] == "osm_id")) {
+            out <- out[-1] # remove osm_id. Not always present
+        }
+        out
+    })
+    keys <- sort (unique (unlist(keysL)))
 
-    df <- lapply (res [1:3], function (i) {
+    tags <- mapply (function (i, k) {
+        i <- i[, k] # remove osm_id column if exists
```
mpadge, line 805:
```suggestion
        i <- i[, k, drop = FALSE] # remove osm_id column if exists
```

mpadge:
Pretty sure this is the problem:
https://github.com/ropensci/osmdata/pull/303/files#r1091737720
Hopefully tests will pass once you accept that suggestion :crossed_fingers:

jmaspons:
indeed, that was the problem. All green now, @mpadge

mpadge:
Cool, thanks. Great that the tests now pass. Let me know when you've updated the news, and we'll merge away ...

PR #309, opened 2023-02-06
Move tests' dependencies to setup.R

jmaspons:
As per `?testthat::test_local()`, move loading of dependencies and common parameters to `setup.R`. With the current code, `testthat::test_local()` or the shortcut Ctrl+Shift+T in Rstudio fail (Error: object 'with_mock_dir' not found) and the tests can only be run manually or doing a full R CMD check, which is much slower. Alternatively, `library(httptest2)` can be added for the files that need it or the namespace scope can be added to each call (`httptest2::with_mock_dir`)

Also reactivate some tests in test-osmdata.R and run all tests except for R-CMD-check in gh workflow (no changes in gh workflows)


mpadge (approved):
That's a great improvement - thanks!

PR #311, opened 2023-02-08
Refresh `README.md` & `codemeta.json`

jmaspons:


mpadge (approved):
Thanks!! Don't worry too much about codemeta - I use a [bash shortcut for that](https://github.com/mpadge/dotfiles/blob/633d0012c9e65f3e8c4423625d98da152c0e081b/.bash_aliases#L34), and run it pretty regularly - but thanks anyway:+1:

PR #316, opened 2023-03-03
Add lat and lon columns for `out * center;` queries in `osmdata_data_frame`

### LINKED ISSUE #315: [FEATURE] Parse latitude and longitude in results from `out * center;` queries
Make it work with `osmdata_data_frame`
``` r
q <- getbb ("Franja de Ponent", featuretype = "relation") %>%
    opq (bb, out = "tags center") %>%
    add_osm_feature ("amenity", "community_center")
res <- osmdata_data_frame(q)
```
`res` should contain columns with the latitude and longitude extracted from the xml result.

A simplified xml result looks like:

``` xml
<osm version="0.6" generator="Overpass API 0.7.59.2 0994154d">
  <note>The data included in this document is from www.openstreetmap.org. The data is made available under ODbL.</note>
  <meta osm_base="2023-03-02T17:46:46Z"/>
  <node id="4282036572" lat="41.1169240" lon="1.2572857">
    <tag k="amenity" v="community_centre"/>
    <tag k="name" v="Casal Popular Sageta de Foc"/>
  </node>
  <way id="311179562">
    <center lat="41.1562828" lon="1.1097734"/>
    <tag k="amenity" v="community_centre"/>
    <tag k="building" v="civic"/>
    <tag k="name" v="Casal Despertaferro"/>
  </way>
  <relation id="5416254">
    <center lat="42.1347037" lon="-0.4061089"/>
    <tag k="amenity" v="community_centre"/>
    <tag k="building" v="yes"/>
    <tag k="community_centre:for" v="juvenile"/>
    <tag k="name" v="Espacio ZETA"/>
  </relation>
</osm>
```

jmaspons:
For calls to `osmdata_data_frame` that do not include the query and the results only contain nodes, there is no way to distinguish if center lat & lon columns have to be added or not. The current implementation adds osm_center_lat/lon in these cases

Fix #315


mpadge (changes requested):
This is another great advance, thanks so much. This review requests just a minor change, but I'm going to put some more major stuff into a separate PR into your local branch, and will comment more there.

tests/testthat/test-data_frame-osm.R
```diff
@@ -221,6 +221,34 @@ test_that ("date", {
...
+        "addr:housenumber", "addr:postcode", "addr:street", "amenity",
+        "building", "building:colour", "building:levels", "building:material",
+        "community_centre", "community_centre:for", "name", "name:ca",
+        "old_name", "operator", "roof:material", "roof:shape", "sport", "type",
+        "website", "wikidata", "wikipedia"
+    )
+    expect_named (x, cols)
+    expect_named (x_no_call, cols)
+    expect_s3_class (x, "data.frame")
+    expect_s3_class (x_no_call, "data.frame")
```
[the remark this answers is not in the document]
jmaspons: Ok, thanks! I will skip the last two tests, because in some cases can be false (I think). `center` returns the coordinates of the center of the bounding box of the object, and I think overpass return the object even if some parts fall outside the bounding box of the query.

mpadge:
> Thanks for the review, @mpadge. Everything addressed, so I'm going to merge

@jmaspons Please don't do so until I have another look through and re-review this. Can you please request another review from me? Thanks

R/get-osmdata.R
```diff
@@ -296,3 +296,23 @@ get_meta_from_cpp_output <- function (res, what = "points") {
...
+#' convert to df, and return only columns with data.
+#'
+#' The "center" components returns from `rcpp_osmdata_df()` are all named with
+#' underscore prefixes. These are prepended here with "osm_center" to provide
+#' standardised names.
+#' @noRd
+get_center_from_cpp_output <- function (res, what = "points") {
+
+    this <- res [[paste0 (what, "_center")]]
+    has_data <- apply (this, 2, function (i) any (nzchar (i)))
```
mpadge, line 311:
```suggestion
    has_data <- apply (this, 2, function (i) any (!is.na (i)))
```

R/get-osmdata.R
```diff
@@ -296,3 +296,23 @@ get_meta_from_cpp_output <- function (res, what = "points") {
...
+    this <- this [, which (has_data), drop = FALSE]
+    if (ncol (this) > 0L) {
+        colnames (this) <- paste0 ("osm_center", colnames (this))
+    }
```
mpadge, line 315:
```suggestion
    } else {
        this <- this [-seq_len (nrow (this)), ]
    }
```

mpadge: That's necessary because of the rownames. Having those means causes rows to be retained even when all columns are removed, so `nrow(this)` remains at whatever value it was, and not zero. That line removes all rows any time that the columns are removed, ensuring a null data.frame is returned.

mpadge (changes requested):
I think that should be okay once those changes have been implemented, but maybe just check downstream in the general workflow here.  One thing i'm not sure about - due to me not understanding it fully :smirk: - is these lines (in "get-osmdata-df.R"::`xml_to_df()`):

``` r
    if (any (missing_center)) { # not a "out * center;" query
        center <- lapply (center, function (x) {
            x[, c ("osm_center_lat", "osm_center_lon")] <- NULL
            x
        })
    }
```
They now don't do anything at all, right? And so can be removed, right?

R/get-osmdata.R
jmaspons: But we need that nrows match other data.frames, so a nrows x 0 dimensions data.frame is fine and necessary, otherwise the code will fail around https://github.com/ropensci/osmdata/blob/31b9a6e95cdd4ab0bb556f7279b73a4c18702983/R/get-osmdata-df.R#L126

And all the tests pass so I think it's fine as it is

jmaspons:
> One thing i'm not sure about - due to me not understanding it fully smirk - is these lines (in "get-osmdata-df.R"::`xml_to_df()`):
> 
> ```r
>     if (any (missing_center)) { # not a "out * center;" query
>         center <- lapply (center, function (x) {
>             x[, c ("osm_center_lat", "osm_center_lon")] <- NULL
>             x
>         })
>     }
> ```
> 
> They now don't do anything at all, right? And so can be removed, right?

This part remove osm_center_* columns if in `doc` exist objects types without center data (checked 3 lines above these lines). The porpoise is to identify queries which should not return center columns (no `out * center;`). I do like this to avoid passing the query to the function. The problematic case is for results that only contain nodes, because nodes always contain lat & lon (relations and ways only when query has `out * center;`).

PR #317, opened 2023-03-04
Fix queries with !match_case and only one value

jmaspons:
The filter requires a ~ bind operator instead of =


mpadge (approved):
Thanks - can you please just update the NEWS with this as a "minor change", and then feel free to merge yourself.

PR #320, opened 2023-03-14
Add parameters from `oqp` to `opq_osm_id`

jmaspons:

PR #321, opened 2023-03-14
 Implement `out:csv` queries compatible with all `overpass_queries`

jmaspons:


mpadge (changes requested):
This is one I've been wanting to do for ages now, so thank you so much for getting on to it! Unfortunatelyu, it doesn't seem to work in current form. This is the example from `opq_csv()`:

``` r
devtools::load_all (".", export_all = TRUE)
#> ℹ Loading osmdata
q <- getbb ("Catalan Countries", format_out = "osm_type_id") %>%
    opq (out = "tags center", osm_type = "relation", timeout = 100) %>%
    add_osm_feature ("admin_level", "7") %>%
    add_osm_feature ("boundary", "administrative") %>%
    opq_csv (fields = c("name", "::type", "::id", "::lat", "::lon"))
comarques <- osmdata_data_frame (q) # without timeout parameter, 0 rows
#> Error in `check_content_type()`:
#> ! Unexpected content type 'text/csv'
#> Expecting one of 'application/xml', 'text/xml'
#> Or suffix '+xml'
#> ℹ Override check with `check_type = FALSE`

#> Backtrace:
#>     ▆
#>  1. └─osmdata::osmdata_data_frame(q)
#>  2.   └─osmdata:::fill_overpass_data(obj, doc, quiet = quiet) at osmdata/R/get-osmdata-df.R:57:4
#>  3.     └─osmdata:::overpass_query(...) at osmdata/R/get-osmdata.R:143:8
#>  4.       └─httr2::resp_body_xml(resp) at osmdata/R/overpass-query.R:229:4
#>  5.         └─httr2:::check_content_type(...)
#>  6.           └─rlang::abort(...)

qid<- opq_osm_id (
    type = "relation",
    id = c ("341530", "1809102", "1664395", "343124")
) %>%
    opq_csv (fields = c ("name", "name:ca"))
cities <- osmdata_data_frame (qid)
#> Error in `check_content_type()`:
#> ! Unexpected content type 'text/csv'
#> Expecting one of 'application/xml', 'text/xml'
#> Or suffix '+xml'
#> ℹ Override check with `check_type = FALSE`

#> Backtrace:
#>     ▆
#>  1. └─osmdata::osmdata_data_frame(qid)
#>  2.   └─osmdata:::fill_overpass_data(obj, doc, quiet = quiet) at osmdata/R/get-osmdata-df.R:57:4
#>  3.     └─osmdata:::overpass_query(...) at osmdata/R/get-osmdata.R:143:8
#>  4.       └─httr2::resp_body_xml(resp) at osmdata/R/overpass-query.R:229:4
#>  5.         └─httr2:::check_content_type(...)
#>  6.           └─rlang::abort(...)
```

<sup>Created on 2023-03-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Plus the other thing that this will absolutely need is a separate function to detect any "out:csv" queries, and automatically stop them being submitted to any other `osmdata_...` functions, with appropriate messages that these may only be submitted to `osmdata_data_frame()`.

---


The errors simply need a change to this line:
https://github.com/ropensci/osmdata/blob/8f3efa566d91ed1ccfc0171a33c57211e210a472/R/overpass-query.R#L229
So that if the body type of the `httr2` return value is "text/csv", then it should be passed to a different function using `httr::resp_body_string()`. The trick is that that returns a single string which then has to be manually shaped into the appropriate number of columns as defined in the query, and equal to the number of items specified in between the brackets of `out:csv(...)`. The result is then also not actually comma delimited, but tab-delimited, and parsing the single character string is going to be a bit tricky...

My attempts looked something like this:
``` r
   doc <- httr2::resp_body_string (resp)
    if (length (doc) == 0L) {
        return (doc)
    }
    doc <- strsplit (doc, "\n|\t") [[1]]
    # This presumes no embedded "()" within csv items!!
    csv_items <- regmatches (query, regexpr ("\\([^\\)]*\\)", query))
    # And this presumes the pattern '",' only occurs between items, not
    # within any!!
    csv_items <- strsplit (gsub ("^\\(|\\)$", "", csv_items), "\\\",") [[1]]
    ncols <- length (csv_items)
    nrows <- length (doc) / ncols
    doc <- t (matrix (doc, nrow = ncols, ncol = nrows))
```

Finally, that `doc` matrix needs to be convrted to a data.frame, using the first row as names. Something along those lines is needed for this to work. Thanks :+1:

jmaspons:
Ouch, right. I have another commit for the `osmdata_*` part. I will add this part to this PR too.

jmaspons:
It would be good to merge #320 to be able to use `out tags` in the examples and tests to avoid recursion and avoid objects with empty tags in the results.

jmaspons:
Precommit hooks propose many style fixes. I will leave them for later to reduce the noise in the PR's diff

mpadge:
> Precommit hooks propose many style fixes. I will leave them for later to reduce the noise in the PR's diff

Yes indeed - please don't commit those unimportant changes in with all of your important stuff here. Thanks

jmaspons:
Tested with more than 60,000 objects and no problem found after the last fix that allow `#` in tag values as some URLs has.

mpadge (approved):
Thanks so much for all this work @jmaspons. Everything works perfectly now! Feel free to merge whenever you're ready :partying_face:

PR #323, opened 2023-04-05
update 'post_process_tags' in 'test-features' to reduce mock file sizes

mpadge:


jmaspons:
New code for me, but as I understand, there are two kinds of tables for feature tags: 1) with `<div class="taglist"`, and 2) with `<table class="wikitable"`.

The tests get a sample of each kind of data to generate the mock call response, dropping the header and all surrounding html code. This MR selects a sample of the rows for the type 2 tables to reduce even more the size of the mock files. Then it tests for `buildind`, a type 2) table. I would add a test for a feature in the type 1) tables (eg. `aerialway`).

For the row's reduction code, perhaps selecting all nodes but the 10 rows of building data + headers  + a sample of type 1) tables, and then using `xml2::xml_remove` could work, but I don't get how that work and should be tried. I'll try later

mpadge:
Thanks! And yes, indeed ...
> For the row's reduction code, perhaps selecting all nodes but the 10 rows of building data + headers + a sample of type 1) tables, and then using xml2::xml_remove could work, but I don't get how that work and should be tried. I'll try later

That is what *should* work, but i can only get `xml_remove()` to remove everything or nothing, not a sub-set.

jmaspons (approved):
Good to go after #324

PR #349, opened 2024-10-28
Document the use of the input query as character strings

jmaspons:


R/get-osmdata-df.R
```diff
@@ -3,9 +3,9 @@
 #'
 #' @inheritParams osmdata_sp
 #' @param q An object of class `overpass_query` constructed with
-#'      \link{opq} and \link{add_osm_feature}. May be be omitted,
-#'      in which case the attributes of the \link{data.frame} will not include
-#'      the query.
+#'      \link{opq} and \link{add_osm_feature} or a string with a valid query.
```
mpadge, line 6: Thanks @jmaspons, this is definitely important. I'd also suggest it should be expanded to:
> a string with a valid query, such as ... \<simple example\>.

Then add another sentence at the end of description for `q`:
> See examples below.

... And then extend the @examples with good examples of all kinds of usage.

jmaspons: Ok. I added examples of complex queries which are not possible to build with regular functions, @mpadge

mpadge (approved):
That's great, thanks!

PR #352, opened 2025-01-31
Depend on httr2 1.1.0

### LINKED ISSUE #351: Error: 'curl_parse_url' is not an exported object from 'namespace:curl' - curl version
Hi all,

This is just a heads up about an error.

I was using osmdata with a previous curl version on my pc (5.2.3 - not 100% sure about last digit), and any operation such as `opq()` or simply `set_overpass_url()` would throw the error:

```r
> osmdata::set_overpass_url("https://maps.mail.ru/osm/tools/overpass/api/interpreter")
Error: 'curl_parse_url' is not an exported object from 'namespace:curl'
```

updating curl to version 6.0.1 seems to solve that.

Maybe useful to make it explicit here:
https://github.com/ropensci/osmdata/blob/bd4fd9cd31406ecbcea74149ff347459f48e54ce/DESCRIPTION#L34

But honestly I didn't make other experiments, sorry.

Thanks

jmaspons:
Fix #351

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -32,7 +32,7 @@ Depends:
     R (>= 3.2.4)
 Imports:
     curl,
-    httr2,
+    httr2 (>= 1.1.0),
     lubridate,
     magrittr,
     methods,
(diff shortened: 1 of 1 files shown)


mpadge:
Closing for these reasons: https://github.com/ropensci/osmdata/issues/351#issuecomment-2628990527

PR #361, opened 2025-06-10
Remove magrittr from imports

### LINKED ISSUE #358: [FEATURE] Reduce osmdata dependencies: magritrr
The base pipe operator `|>` is available since R 4.1 and can substitute magrittr::`%>%`(). The use of `%>%` suppose a tiny overhead but can be avoided internally without affecting the API. From the user perspective, removing `magrittr` from imports can break some code because it would require explicitly loading the package if the user use `%>%` and some examples in `osmdata` would need updates.

We have different options with different:

- Require R >= 4.1 and use `|>` instead of `%>%`.
- Avoid `|>` and `%>%` in the internal code.
  - `%>%` can be keep it exported. `magrittr` can be removed from dependencies in the future without currently impacting the users.
  - Rewrite the examples and vignettes without pipes. Remove `magrittr` from the dependencies and remove `%>%` from exported functions.

What do you think, @mpadge ?

jmaspons:
User code relaying on reexported pipe `%>%` from `osmdata` must explicitly load it with `library(magrittr)`.  Code examples, tests, and vignettes now use the pipe from base (`|>`) available since R 4.1. Using older versions will work, but the code from some examples and vignettes will need adjustments to run and is not supported.

Apparently, we don't break any package as tested with revdep and the tests with devtools::check_win_oldrelease() use versions already supporting the base pipe.

FIX #358


R/getbb.R
```diff
@@ -375,10 +375,8 @@ get_geotext_poly <- function (obj) {
     gt_p <- NULL
     indx <- which (!(seq (nrow (obj)) %in% indx_multi))
     index_final <- index_final [indx]
-    gt_p <- obj$geotext [indx] %>%
-        gsub ("POLYGON\\(\\(", "", .) %>%
-        gsub ("\\)\\)", "", .) %>%
-        strsplit (split = ",")
+    gt_p <- gsub ("(POLYGON\\(\\()|(\\)\\))", "", obj$geotext [indx])
```
mpadge, line 378:
```suggestion
    ptn <- "(POLYGON\\(\\()|(\\)\\))"
    gt_p <- gsub (ptn, "", obj$geotext [indx])
```
(just to make the line shorter, and a bit easier to read)

PR #367, opened 2025-06-12
Split package and osmdata-class docs

jmaspons:
See current doc at https://docs.ropensci.org/osmdata/reference/osmdata.html


mpadge (approved):
Oh, great catch! Thank you so much!

PR #369, opened 2025-06-13
Use @examplesIf to show examples with or without pipes according to R version

jmaspons:
Users with R < 4.1 don't see pipes as they are not supported. Web version should show pipes as pkgdown runner use R 4.5

PR #371, opened 2025-06-16
Depend on R >= 4.1 after dropping magrittr dependence

### LINKED ISSUE #358: [FEATURE] Reduce osmdata dependencies: magritrr

jmaspons:
Simplify examples assuming R >= 4.1 + fixes from examplesIf branch

FIX #358
DROP #369

### DIFF
--- a/R/elevation.R
+++ b/R/elevation.R
@@ -12,25 +12,17 @@
 #' @return A modified version of the input `dat` with an additional `z_` column
 #' appended to the vertices.
 #' @family transform
+#'
 #' @examples
-#' # Bounding box of "omaha nebraska":
-#' bb <- c (-96.2682, 41.190585, -95.871141, 41.3930618)
-#' query <- opq (bb)
-#' query <- add_osm_feature (query, key = "highway")
-#' # Equivalent to:
 #' \dontrun{
 #' query <- opq ("omaha nebraska") |>
 #'     add_osm_feature (key = "highway")
-#' }
-#' # Elevation can only be app;lied to \pkg{silicate} 'SC'-class data:
-#' \dontrun{
+#' # Elevation can only be applied to \pkg{silicate} 'SC'-class data:
 #' dat <- osmdata_sc (query)
 #' dat$vertex
-#' }
 #' # The vertex table will have columns ("x_", "y_", "vertex_"). Then
 #' # download elevation data from \url{https://srtm.csi.cgiar.org/srtmdata/}
 #' # (or elsewhere), and add elevation column, "z_" with:
-#' \dontrun{
 #' dat <- osm_elevation (dat, elev_file = "/path/to/elevation/data.tiff")
 #' }
 #' @export
--- a/R/get-osmdata-df.R
+++ b/R/get-osmdata-df.R
@@ -21,17 +21,10 @@
 #' @export
 #'
 #' @examples
-#' # Bounding box of "hampi india":
-#' bb <- c (76.4410201, 15.3158, 76.4810201, 15.3558)
-#' query <- opq (bb)
-#' query <- add_osm_feature (query, key = "historic", value = "ruins")
-#' # Equivalent to:
 #' \dontrun{
 #' query <- opq ("hampi india") |>
 #'     add_osm_feature (key = "historic", value = "ruins")
-#' }
 #' # Then extract data from 'Overpass' API
-#' \dontrun{
 #' hampi_df <- osmdata_data_frame (query)
 #' attr (hampi_df, "bbox")
 #' attr (hampi_df, "overpass_call")
--- a/R/get-osmdata-sf.R
+++ b/R/get-osmdata-sf.R
@@ -15,18 +15,11 @@
 #' @export
 #'
 #' @examples
-#' # Bounding box of "hampi india":
-#' bb <- c (76.4410201, 15.3158, 76.4810201, 15.3558)
-#' query <- opq (bb)
-#' query <- add_osm_feature (query, key = "historic", value = "ruins")
-#' # Equivalent to:
 #' \dontrun{
 #' query <- opq ("hampi india") |>
 #'     add_osm_feature (key = "historic", value = "ruins")
-#' }
 #' # Then extract data from 'Overpass' API
-#' \dontrun{
-#' hampi_sf <- osmdata_sf ()
+#' hampi_sf <- osmdata_sf (query)
 #' }
 #'
 #' # Complex query as a string (not possible with regular osmdata functions)
--- a/R/get-osmdata-sp.R
+++ b/R/get-osmdata-sp.R
@@ -19,18 +19,11 @@
 #' @export
 #'
 #' @examples
-#' # Bounding box of "hampi india":
-#' bb <- c (76.4410201, 15.3158, 76.4810201, 15.3558)
-#' query <- opq (bb)
-#' query <- add_osm_feature (query, key = "historic", value = "ruins")
-#' # Equivalent to:
 #' \dontrun{
 #' query <- opq ("hampi india") |>
 #'     add_osm_feature (key = "historic", value = "ruins")
-#' }
 #' # Then extract data from 'Overpass' API
-#' \dontrun{
-#' hampi_sp <- osmdata_sp ()
+#' hampi_sp <- osmdata_sp (query)
 #' }
 #'
 #' # Complex query as a string (not possible with regular osmdata functions)
--- a/R/get-osmdata-xml.R
+++ b/R/get-osmdata-xml.R
@@ -20,18 +20,11 @@
 #' @export
 #'
 #' @examples
-#' # Bounding box of "hampi india":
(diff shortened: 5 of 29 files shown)


R/elevation.R
```diff
@@ -12,17 +12,13 @@
...
+#'
 #' @examples
 #' # Bounding box of "omaha nebraska":
 #' bb <- c (-96.2682, 41.190585, -95.871141, 41.3930618)
-#' query <- opq (bb)
-#' query <- add_osm_feature (query, key = "highway")
-#' # Equivalent to:
-#' \dontrun{
-#' query <- opq ("omaha nebraska") |>
+#' query <- opq (bb) |>
```
mpadge, line 19:
```suggestion
#' query <- opq ("omaha nebraska") |>
```
I think this, and all others, are clearer if the not-run parts just have the place names, because that's how I imagine this function mostly gets used (including by me).

R/get-osmdata-df.R
```diff
@@ -23,13 +23,8 @@
 #' @examples
 #' # Bounding box of "hampi india":
 #' bb <- c (76.4410201, 15.3158, 76.4810201, 15.3558)
-#' query <- opq (bb)
-#' query <- add_osm_feature (query, key = "historic", value = "ruins")
-#' # Equivalent to:
-#' \dontrun{
-#' query <- opq ("hampi india") |>
+#' query <- opq (bb) |>
```
mpadge, line 26:
```suggestion
#' query <- opq ("hampi india") |>
```

R/get-osmdata-sf.R
```diff
@@ -17,16 +17,11 @@
 #' @examples
 #' # Bounding box of "hampi india":
 #' bb <- c (76.4410201, 15.3158, 76.4810201, 15.3558)
-#' query <- opq (bb)
-#' query <- add_osm_feature (query, key = "historic", value = "ruins")
-#' # Equivalent to:
-#' \dontrun{
-#' query <- opq ("hampi india") |>
+#' query <- opq (bb) |>
```
mpadge, line 20:
```suggestion
#' query <- opq ("hampi india") |>
```

R/get-osmdata-sp.R
```diff
@@ -21,16 +21,11 @@
 #' @examples
 #' # Bounding box of "hampi india":
 #' bb <- c (76.4410201, 15.3158, 76.4810201, 15.3558)
-#' query <- opq (bb)
-#' query <- add_osm_feature (query, key = "historic", value = "ruins")
-#' # Equivalent to:
-#' \dontrun{
-#' query <- opq ("hampi india") |>
+#' query <- opq (bb) |>
```
mpadge, line 24:
```suggestion
#' query <- opq ("hampi india") |>
```

R/get-osmdata-xml.R
```diff
@@ -22,16 +22,11 @@
 #' @examples
 #' # Bounding box of "hampi india":
 #' bb <- c (76.4410201, 15.3158, 76.4810201, 15.3558)
-#' query <- opq (bb)
-#' query <- add_osm_feature (query, key = "historic", value = "ruins")
-#' # Equivalent to:
-#' \dontrun{
-#' query <- opq ("hampi india") |>
+#' query <- opq (bb) |>
```
mpadge, line 25:
```suggestion
#' query <- opq ("hampi india") |>
```

R/get-osmdata-df.R
```diff
@@ -23,13 +23,8 @@
 #' @examples
 #' # Bounding box of "hampi india":
 #' bb <- c (76.4410201, 15.3158, 76.4810201, 15.3558)
```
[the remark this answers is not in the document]
jmaspons: Done, @mpadge ! I think it's ready to merge

jmaspons:
Ready for a new round, @mpadge

For https://github.com/ropensci-review-tools/pkgcheck/pull/246#issuecomment-2999943743, what's the problem running the tests? Is just the long runtime? Errors if !has_internet()?

R/get-osmdata-df.R
```diff
@@ -21,17 +21,11 @@
 #' @export
 #'
 #' @examples
-#' # Bounding box of "hampi india":
-#' bb <- c (76.4410201, 15.3158, 76.4810201, 15.3558)
-#' query <- opq (bb)
-#' query <- add_osm_feature (query, key = "historic", value = "ruins")
-#' # Equivalent to:
 #' \dontrun{
+#' # Bounding box of "hampi india":
```
mpadge, line 25:
```suggestion
```

R/opq.R
```diff
@@ -310,9 +310,9 @@ paste_features <- function (key, value, key_pre = "", bind = "=",
 #'     add_osm_feature (key = "highway", value = "!primary")
 #'
 #' # key negation without warnings
-#' q3 <- opq ("Vinçà", osm_type="node") |>
-#'     add_osm_feature (key = c("name", "!name:ca"))
-#' q4 <- opq ("el Carxe", osm_type="node") |>
+#' q3 <- opq ("Vinçà", osm_type = "node") |>
```
mpadge, line 313: Nice, thanks for finding these!! :+1:

R/poly2line.R
```diff
@@ -14,22 +14,14 @@
 #'
 #' @family transform
 #' @examples
-#' # Bounding box of "Colchester UK":
-#' bb <- c (0.6993788, 51.7657055, 1.026803, 51.977153)
-#' query <- opq (bb)
-#' query <- add_osm_feature (query, key = "highway")
-#' # Equivalent to:
 #' \dontrun{
+#' # Bounding box of "Colchester UK":
```
mpadge, line 18:
```suggestion
```

R/utils.R
```diff
@@ -17,22 +17,15 @@ is_datetime <- function (x) {
...
 #' @family transform
+#'
 #' @examples
-#' # Bounding box of "hampi india":
-#' bb <- c (76.4410201, 15.3158, 76.4810201, 15.3558)
-#' query <- opq (bb)
-#' query <- add_osm_feature (query, key = "historic", value = "ruins")
-#' # Equivalent to:
 #' \dontrun{
+#' # Bounding box of "hampi india":
```
mpadge, line 23:
```suggestion
```

mpadge (approved):

R/opq.R
jmaspons: The merit goes for pre-commit & styler 😋

mpadge:
> > I was just referring to the fact that in our case we have to have examples witth `\dontrun`.
> 
> But why? It's because it would take too much time to check the package? You don't want to bother overpass servers? Instabilities due to relaying on overpass servers? I tried to [check all the code in examples](https://github.com/ropensci/osmdata/actions/runs/15858761034/job/44710456194) but the donttest container from r-hub seems to skip them anyway [r-hub/containers#94](https://github.com/r-hub/containers/issues/94)

Oh sorry, no just because `opq("place name")` calls the Nominatim server, and so will fail CRAN's requirement for "graceful failure" if the server is not reachable (which happens often enough). So they can't be run under any `@examplesIf` condition (except maybe [`examplesIf ami::on_cran()`](https://briandconnelly.github.io/ami/reference/cran.html), but I doubt they'd take kindly to that :smirk:). Everything is tested anyway, so it's fine, but I do think this is a clear case where simple `\dontrun{}` statements like we have make for the clearest code in our examples.

> By the way, [valgrind seems to find problems](https://github.com/ropensci/osmdata/actions/runs/15858761034/job/44710456231). I don't know if they are relevant or fixable

Oh thanks, always good to check those, but they're all caused by stuff outside our package. The key there is they're only "possibly lost" at the very end. That's generally okay, while "definitely lost" signal genuine problems. Our own https://github.com/ropensci/osmdata/blob/main/tests/valgrind-test.R is run each time with tests anyway, and that should be considered the definitive checker of valgrind errors. (And in case you're unsure, valgrind checks for memory leaks in compiled code.)

jmaspons:
Ok! It seems the valgrind test is not ran anymore and tests/valgrind-script.R was removed, but now I see that the logs from r-hub points to libudunits2 and libexpat.

mpadge:
Oh yeah, you're right! The function is just a definition. The script was the bit that actually ran it. I'll try to get it going again. Thanks!

Issue #374, opened 2025-07-04
[BUG/QUESTION] Classes returned by osmdata_sf and osmdata_sc


agila5:
Hi again! This is probably not so important, but I find it quite weird that the classes returned by `osmdata_sf()` and `osmdata_sc()` (which, based on my understanding, correspond to different representations of the same underlying data) are so different: 

``` r
library(osmdata)
#> Data (c) OpenStreetMap contributors, ODbL 1.0. https://www.openstreetmap.org/copyright

opq("hampi india") |> 
  add_osm_feature (key = "historic", value = "ruins") |> 
  osmdata_sf() |> 
  class()
#> [1] "list"       "osmdata"    "osmdata_sf"

opq("hampi india") |> 
  add_osm_feature (key = "historic", value = "ruins") |> 
  osmdata_sc() |> 
  class()
#> [1] "SC"         "sc"         "osmdata_sc"
```

<sup>Created on 2025-07-04 with [reprex v2.1.1.9000](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> Warning in system2("quarto", "-V", stdout = TRUE, env = paste0("TMPDIR=", :
#> running command '"quarto"
#> TMPDIR=C:/Users/andre/AppData/Local/Temp/Rtmpe6o5rp/file53054c6478e -V' had
#> status 1
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.4.2 (2024-10-31 ucrt)
#>  os       Windows 11 x64 (build 26100)
#>  system   x86_64, mingw32
#>  ui       RTerm
#>  language (EN)
#>  collate  Italian_Italy.utf8
#>  ctype    Italian_Italy.utf8
#>  tz       Europe/Rome
#>  date     2025-07-04
#>  pandoc   3.2 @ C:/Program Files/RStudio/resources/app/bin/quarto/bin/tools/ (via rmarkdown)
#>  quarto   NA @ C:\\Users\\andre\\AppData\\Local\\Programs\\Quarto\\bin\\quarto.exe
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version    date (UTC) lib source
#>  cli           3.6.4      2025-02-13 [1] CRAN (R 4.4.3)
#>  curl          6.2.1      2025-02-19 [1] CRAN (R 4.4.2)
[...]

PR #376, opened 2025-07-07
update sc class def to  #374

mpadge:


agila5:
LGTM but, to be honest, I don't understand the benefits of preserving the `c("SC", "sc")` classes returned by `silicate`. Is there any?

Issue #377, opened 2025-07-10
[BUG] Unclear error message with getbb


agila5:
See the reprex below

``` r
osmdata::getbb("Milan, Italy", format_out = "sf_polygon")
#> Error in names(gt_mp) <- as.character(index_final): 'names' attribute [2] must be the same length as the vector [1]
```

<sup>Created on 2025-07-10 with [reprex v2.1.1.9000](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> Warning in system2("quarto", "-V", stdout = TRUE, env = paste0("TMPDIR=", :
#> running command '"quarto"
#> TMPDIR=C:/Users/andre/AppData/Local/Temp/RtmpsRWZXz/file1943adb5ac2 -V' had
#> status 1
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.4.2 (2024-10-31 ucrt)
#>  os       Windows 11 x64 (build 26100)
#>  system   x86_64, mingw32
#>  ui       RTerm
#>  language (EN)
#>  collate  Italian_Italy.utf8
#>  ctype    Italian_Italy.utf8
#>  tz       Europe/Rome
#>  date     2025-07-10
#>  pandoc   3.2 @ C:/Program Files/RStudio/resources/app/bin/quarto/bin/tools/ (via rmarkdown)
#>  quarto   NA @ C:\\Users\\andre\\AppData\\Local\\Programs\\Quarto\\bin\\quarto.exe
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version    date (UTC) lib source
#>  cli           3.6.4      2025-02-13 [1] CRAN (R 4.4.3)
#>  curl          6.2.1      2025-02-19 [1] CRAN (R 4.4.2)
#>  digest        0.6.37     2024-08-19 [1] CRAN (R 4.4.2)
#>  evaluate      1.0.3      2025-01-10 [1] CRAN (R 4.4.2)
#>  fastmap       1.2.0      2024-05-15 [1] CRAN (R 4.4.2)
#>  fs            1.6.5      2024-10-30 [1] CRAN (R 4.4.2)
#>  glue          1.8.0      2024-09-30 [1] CRAN (R 4.4.2)
#>  htmltools     0.5.8.1    2024-04-04 [1] CRAN (R 4.4.2)
[...]

PR #378, opened 2025-07-10
start fix getbb sf_polygon for #377

mpadge:


mpadge:
Thanks @jmaspons. I don't exactly recall what I was trying to do with this PR - I think it was just an attempt to simplify the code. But also note [the comment I made here](https://github.com/ropensci/osmdata/issues/377#issuecomment-3199913574), which here likely just means that any returned multipolygon objects may not strictly be "correct" in `sf` sense, even thought correct in OSM sense. Maybe just check with the [example you gave there](https://github.com/ropensci/osmdata/issues/377#issuecomment-3198237776)? Anyway, I'll flag PR ready for review, and you can modify as you like :+1:

jmaspons (commented):
Now, `format_out = "polygon"` and `format_out = "sf_polygon"` works properly. I only have doubts about returning all polygons or just the first one in the results from nominatim documentation says (but not sure if it was the case, even before this PR). I would like to be sure that other functions expecting these results as input can deal with it properly. Thoughts on this, @mpadge ?

mpadge:
> That's not a rectangular bounding box any more. Is a rectangle the expected bb for points and Accra was a point years ago, @mpadge ? We should find another example and add tests

Yes indeed, that was how things used to be. I'd say it'll be better to just remove that example, because any we do happen to find will likely also only be temporary, until somebody defines a proper polygon anyway. Just add some text to suggest that polygons may be rectangles if not otherwise defined in OSM itself?

jmaspons:
Fixed and tested the case with no polygons for sf and removed the example. Not documented, but a message is printed in these cases.

 Also, I tried to improve the docs and remove a part that I think didn't match the behaviour of the `getbb()`.

All checks in green and happy from my side. If there are no news, I will merge in a day. Then, I will restart the release.

PR #391, opened 2025-09-03
Use functions from R base instead of methods

jmaspons:
From `?methods::is`:
"Although inherits is defined for S3 classes, it has been modified so that the result returned is nearly always equivalent to is, both for S4 and non-S4 objects. Since it is implemented in C, it is somewhat faster."


jmaspons:
Done!

mpadge (approved):
Approved - merge away!

Issue #392, opened 2025-09-06
[BUG] Uninformative error for empty results in getbb(..., format_out = "data.frame")


jmaspons:
``` r
osmdata::getbb("Antilles Neerlandeses", format_out = "data.frame")
#> Error in obj[, utf8cols]: incorrect number of dimensions
```

<sup>Created on 2025-09-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Before the 0.3.0 release, the function returned `list()` but it's inconsistent with other `format_out` parameters which return:
``` r
osmdata::getbb("Antilles Neerlandeses", format_out = "sf_polygon")
#> Error in osmdata::getbb("Antilles Neerlandeses", format_out = "sf_polygon") : 
#>  `place_name` 'Antilles Neerlandeses' can't be found
```

I think an error for an empty result is too much, but it's good to be consistent. So I would go for returning empty results following the format of the expected output. Thoughts, @mpadge ?

PR #394, opened 2025-09-19
getbb() with empry results

jmaspons:
`gettbb() `throws a warning instead of an error for empty results and return the (empty) result with the expected type according to `format_out` parameter

FIX: #392


mpadge (approved):
That's a great enhancement - thanks so much @jmaspons

PR #406, opened 2025-10-28
Avoid skipping test-unname.R for test-coverage action

jmaspons:
Perhaps skipped because of the workflow name mismatch

https://github.com/ropensci/osmdata/blob/3d551f38ba8d58bb3c1577a20ece25115ec53403/tests/testthat/test-unname.R#L3-L4


jmaspons:
Is there any problem with the complete tests in CRAN? We can skip the extra tests there if needed

PR #408, opened 2025-11-03
mention github as primary site in readme, and list other locations

mpadge:
@jmaspons Would this be okay to add this information to the README? Thanks!


jmaspons (approved):
Good, this clarifies the current situation.

PR #416, opened 2026-03-16
Return a POSIXct timestamp instead of a non-standard locale dependent date

jmaspons:


mpadge (changes requested):
I'll request changes here, just so I see your response to https://github.com/ropensci/osmdata/pull/416#issuecomment-4073036082. Thanks!

mpadge (approved):
Great, thanks!

Issue #255, opened 2021-11-30
[FEATURE] re-design of AND versus OR queries


mpadge:
@Mashin6 raised valid concerns in #252, supported by @stragu, that the current design in counter-intuitive:

**OR**
``` r
add_osm_features (features = c (...))
```

**AND**
``` r
add_osm_feature() |>
    add_osm_feature() |> ...
```

This needs to be improved, for which suggestions are requested!

stragu:
Similar to what @Mashin6 said in the mentioned issue, I believe it would make a lot more sense for newcomers to:

1. build a dataset by "getting" the results of various key-pair combinations, i.e. an OR operation, using `get_osm_features(key = "xxx", value = "xxx") |> get_osm_features(key = "xxx", value = "xxx")`
1. use several key-value pairs that all need to be matched for the returned features, i.e. an AND operation, using `get_osm_features(match_all = c(...))`

The name `get_osm_features()` is chosen so it doesn't conflict with the existing `add_osm_feature()` and `add_osm_features()` functions (which could be deprecated over several osmdata versions, as the new `get_osm_features()` would replace them). The argument `match_all` leaves no space for misinterpretation.

That would fit better the way most people would read the code, replacing the pipe with "then":

1. "get these OSM features, and _then_ also get these OSM features, etc."
2. "Add the OSM features that _match_all_ these key-value pairs."

Adapting Mashin6's hypothetical example, a sample command would look like this:

```r
opq() |>
   get_osm_features(match_all = c("natural = tree", "leaf_type = broadleaved") |>
   get_osm_features(match_all = c("leisure = park", "name ~ public") |>
   get_osm_features(key = "amenity", value = "bench")
```

... which would return an opq object that would retrieve **broad leaved trees, benches, and parks with "public" in their name**.

In the meantime, I'm wondering if aliasing `add_osm_feature()` and `add_osm_features()` with `filter_osm_feature()` and `filter_osm_features()` respectively would help clarifying how they behave, given that even the documentation draws a parallel with `dplyr::filter()` ?

stragu:
I'm also thinking an ellipsis argument could be used instead of a `match_all`, if it is clear enough that all these values will be combined with "AND". You could then simplify the same command with:

```r
opq() |>
   get_osm_features("natural = tree", "leaf_type = broadleaved") |>
   get_osm_features("leisure = park", "name ~ public") |>
   get_osm_features(key = "amenity", value = "bench")
```

mpadge:
[...]

PR #417, opened 2026-03-17
Rename query builders

jmaspons:
Rename `add_osm_feature()` -> `filter_osm_tags()` and `add_osm_features()` -> `add_osm_tags()`. Keep the old alias working and silent for now, the old names will be DEPRECATED in a future version and will throw warnings.

Make docs, parameters and variable names consistent for tags / features use. Feature refers to entities from the real world represented by OSM objects. Tags are the attributes of OSM objects that characterize the features.

Discussed in #255


mpadge (changes requested):
Thanks @jmaspons I'm going to request changes here just so that I'll see when you've re-requested review. The changes I've requested are described in https://github.com/ropensci/osmdata/issues/255#issuecomment-4088660973

PR #426, opened 2026-07-10
Fix points without metadata

jmaspons:
Error reproducible following https://github.com/jeremylfarrell/lakefetch/issues/2#issuecomment-4935272198


mpadge (approved):
Wow, thanks for the quick fix. Can you please drop a note in the issue this came from? Thank you!

Issue #427, opened 2026-07-16
Alternative Overpass URLs possibly outdated?


cforgaci:
Dear maintainer(s) of the awesome `osmdata`,

While checking the `set_overpass_url()` documentation, I noticed that, aside from the default `"https://overpass-api.de/api/interpreter"`, the URLs with global coverage do not match those in the [Overpass API documentation](https://wiki.openstreetmap.org/wiki/Overpass_API).

I came across this issue while being stuck in `"Waiting 60s for retry backoff"` due to the URL being set to the non-default `"https://overpass.kumi.systems/api/interpreter"`, which appears to be known now in the Overpass documentation as `"https://overpass.private.coffee/api/interpreter"`.

Now, my issue is solved since I set the URL back to the default, but I do not remember changing it, so I wonder if there was an automatic shift to alternative URLs that might benefit from an update.

jmaspons:
Good point! We can test the different [public instances](https://wiki.openstreetmap.org/wiki/Overpass_API#Public_Overpass_API_instances) and update the set from which we sample when loading the package. See the current implementation:

https://github.com/ropensci/osmdata/blob/0e3f9c6827ba10f64db97a10ddb96333ffbbc20c/R/zzz.R#L9-L12

cforgaci:
@jmaspons I see that those lines were added in 2020, which is consistent with the URLs available in the Overpass API docs at that time. Besides randomly sampling a public instance, it might help to check something like the URL's status before setting it, as the list may change again in the future. I am happy to give this a try in a PR if it helps.

jmaspons:
Thanks! Go ahead, @cforgaci

mpadge:
(Duplicated comment copied from #428)

Thanks @cforgaci for helping to improve this situation. I suspect the rate-limiting that's happening is also region-specific, so likely to be different in different regions? Whatever the real cause, I think this shows a need to offer more refined and explicit control over overpass URLs. While this PR is good, and we'd have no problems merging it, if you wanted to extend, I think it would be good to enable the following two improvements:

### 1 Auto-identify a currently working Overpass URL

The main issue with current code is that every Overpass call is wrapped in `httr2::req_retry(req, max_tries = 10L)`. That includes
https://github.com/ropensci/osmdata/blob/0e3f9c6827ba10f64db97a10ddb96333ffbbc20c/R/overpass-query.R#L35-L37
[...]

PR #428, opened 2026-07-16
Update URLs of public Overpass API instances

cforgaci:
As agreed in https://github.com/ropensci/osmdata/issues/427, I am opening a PR here with the updated URLs. Besides the default `"https://overpass-api.de/api/interpreter"`, only `"https://maps.mail.ru/osm/tools/overpass/api/interpreter"` seems to work with a usage policy comparable to the default instance.

I also suggested dropping the region-specific instances from the `set_overpass_url()` docs, especially because their usage policies are not specific enough on rate limits. The user can still go to https://wiki.openstreetmap.org/wiki/Overpass_API and get more information about other instances.

I do not know if URL checks should be done in `.onLoad()`, so I am not sure how to make this future-proof. This should work for now, but the issue might arise again if the instances change.

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -8,7 +8,7 @@
     ## Added and edited code here by JimShady to use random API each time.
     available_apis <- c (
         "https://overpass-api.de/api/interpreter",
-        "https://overpass.kumi.systems/api/interpreter"
+        "https://maps.mail.ru/osm/tools/overpass/api/interpreter"
     )
 
     op.osmdata <- list ( # nolint
@@ -64,15 +64,7 @@ get_overpass_url <- function () {
 #' are:
 #' \itemize{
 #' \item "https://overpass-api.de/api/interpreter" (default)
-#' \item "https://overpass.kumi.systems/api/interpreter"
-#' \item "https://overpass.osm.rambler.ru/cgi/interpreter"
-#' \item "https://api.openstreetmap.fr/oapi/interpreter"
-#' \item "https://overpass.osm.vi-di.fr/api/interpreter"
-#' }
-#' Additional APIs with limited local coverage include:
-#' \itemize{
-#' \item "https://overpass.osm.ch/api/interpreter" (Switzerland)
-#' \item "https://overpass.openstreetmap.ie/api/interpreter" (Ireland)
+#' \item "https://maps.mail.ru/osm/tools/overpass/api/interpreter"
 #' }
 #'
 #' For further details, see
--- a/man/set_overpass_url.Rd
+++ b/man/set_overpass_url.Rd
@@ -17,15 +17,7 @@ Set the URL of the specified overpass API. Possible APIs with global coverage
 are:
 \itemize{
 \item "https://overpass-api.de/api/interpreter" (default)
-\item "https://overpass.kumi.systems/api/interpreter"
-\item "https://overpass.osm.rambler.ru/cgi/interpreter"
-\item "https://api.openstreetmap.fr/oapi/interpreter"
-\item "https://overpass.osm.vi-di.fr/api/interpreter"
-}
-Additional APIs with limited local coverage include:
-\itemize{
-\item "https://overpass.osm.ch/api/interpreter" (Switzerland)
-\item "https://overpass.openstreetmap.ie/api/interpreter" (Ireland)
+\item "https://maps.mail.ru/osm/tools/overpass/api/interpreter"
 }
 }
 \details{
--- a/vignettes/osmdata.Rmd
+++ b/vignettes/osmdata.Rmd
@@ -192,7 +192,7 @@ base url:
 
 ```{r}
 get_overpass_url ()
-new_url <- "https://overpass.openstreetmap.ie/api/interpreter"
+new_url <- "https://maps.mail.ru/osm/tools/overpass/api/interpreter"
 ```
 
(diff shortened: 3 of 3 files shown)


mpadge:
Thanks @cforgaci for helping to improve this situation. I suspect the rate-limiting that's happening is also region-specific, so likely to be different in different regions? Whatever the real cause, I think this shows a need to offer more refined and explicit control over overpass URLs. While this PR is good, and we'd have no problems merging it, if you wanted to extend, I think it would be good to enable the following two improvements:

### 1 Auto-identify a currently working Overpass URL

The main issue with current code is that every Overpass call is wrapped in `httr2::req_retry(req, max_tries = 10L)`. That includes
https://github.com/ropensci/osmdata/blob/0e3f9c6827ba10f64db97a10ddb96333ffbbc20c/R/overpass-query.R#L35-L37
so even status queries immediately lead to staged backoffs, and block any futher usage of any function here. That should be largely avoidable with a function like this:
 
``` r
status_check <- function (u) {
    st_type <- "status"
    status_url <- gsub ("interpreter", st_type, u)
    req <- httr2::request (status_url)
    x <- tryCatch (
        httr2::req_perform (req),
        error = function (e) e
    )
    if (inherits (x, "error")) {
        ret <- list (status = x$status, message = x$message)
    } else {
        ret <- list (status = x$status_code, message = "OK") }
    ret
}
```
Calling that on a vector of possible URLs then gives this (with of course no annoying backoff):
``` r
overpass_url <- c (
    "https://overpass.kumi.systems/api/interpreter",
    "https://overpass.private.coffee/api/interpreter",
    "https://overpass-api.de/api/interpreter"
)
lapply (overpass_url, status_check)
#> [[1]]
#> [[1]]$status
#> [1] 429
#> 
#> [[1]]$message
#> [1] "HTTP 429 Too Many Requests."
#> 
#> 
#> [[2]]
#> [[2]]$status
#> [1] 429
#> 
#> [[2]]$message
#> [1] "HTTP 429 Too Many Requests."
#> 
#> 
#> [[3]]
#> [[3]]$status
#> [1] 200
#> 
#> [[3]]$message
#> [1] "OK"
```

<sup>Created on 2026-07-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

That shows that, on my computer (in Germany), I should `set_overpass_url("https://overpass-api.de/api/interpreter")`. Others may see different results, and in cases where more than 1 returns OK, I guess a `sample()` approach like current form would be okay. Main thing is **no retry there**, to enable url to be set without `httr2` backoff.

But it's might not be a great idea to auto-run that in `.onLoad`, or if so, to make sure it's only triggered on `if (interactive())`, and maybe other conditions?

### 2 Allow more direct user control over URLs, and expose possibilities directly to user

The set of URLs are currently not exposed at all, but it's pretty clear that the package would be more usable if people not only had control, but knew what the possible options were. I think it would be good to expose the `overpass_url` vector of possible values directly to users, with a clear message when that function is called that links directly to https://wiki.openstreetmap.org/wiki/Overpass_API#Public_Overpass_API_instances.

We currently have this:
https://github.com/ropensci/osmdata/blob/0e3f9c6827ba10f64db97a10ddb96333ffbbc20c/R/zzz.R#L65-L71

But there should be one single source of truth within this package, and I'm suggesting that should be an exported function. Functions like `set_overpass_url()` can then just have `# @examples list_overpass_urls()` or whatever it's called to show values in the docs there. I have no clear vision whether the auto-setting of URLs should be in an exported or internal function, and will happily what to see where this evolves.

---

## Summary

@cforgaci Let us know whether you might have time and willingness to start implementing these kinds of suggestions. No worries if not; we'll just merge this PR and proceed from there, but it would be cool if you wanted to develop something like this. Thanks!

cforgaci:
@mpadge, @jmaspons, thanks for the detailed feedback on how this can be further improved. I'd like to start implementing this following your suggestions sometime in the following week.

cforgaci:
@jmaspons, @mpadge, can this wait for a while longer? I was planning to pick this up this week, but I could not find the time and I am transitions into my summer break. I should be able to resume when I am back mid-August.

mpadge:
No worries from me. Thanks for letting us know

jmaspons:
That's fine!

mpadge:
@cforgaci Gentle reminder ...

cforgaci:
@mpadge, sorry for the delay with this. I tried to implement the two improvements you suggested. Let me know if this is what you had in mind.

cforgaci:
> @cforgaci That's perfect - thank you so much! Can you please just:
> 
> 1. Update https://github.com/ropensci/osmdata/blob/main/NEWS.md, adding this as a "major change"
> 2. Add yourself as an "aut" to https://github.com/ropensci/osmdata/blob/main/DESCRIPTION ? (Criterion there: Authors = anybody who adds new functionality, which you've definitely done here.)

@mpadge, I did so just now. Thank you.

mpadge:
Great; thanks so much Cladiu!

@jmaspons Hope you don't mind me taking over this one - I presumed you were away on holidays or something, and this really needs to be merged, because I have heard elsewhere that the package without this had become basically unusable for many. That also means we need to get this fix on to CRAN asap, if you've got time. Thanks

jmaspons:
Thanks to booth, good work! I will  try to do a release next week.

PR #430, opened 2026-07-18
Add user agent in requests to servers

### LINKED ISSUE #429: [FEATURE] Add user agent to requests to servers
As talked in https://github.com/ropensci/osmdata/issues/427#issuecomment-5001130200, add https://httr2.r-lib.org/reference/req_user_agent.html with `string = "osmdata"` to the requests to reduce the frequency of servers blocking the queries. At least for `https://overpass-api.de/api/interpreter`, this is expected as explained in https://wiki.openstreetmap.org/wiki/Overpass_API#Public_Overpass_API_instances

Could be set as a package option and allow users to modify it as in osmapiR https://github.com/ropensci/osmapiR/blob/1e70cdabd6c5f20b61335763d335516e49fa8399/R/osmapiR_request.R#L5

jmaspons:
Fix #429


mpadge:
Thanks @jmaspons, this seems like a very sensible thing to add. Can we just have a test (with mock results via httptest2) that tests for responses both with and without the user agent? That will help just in case anything changes down the line.

jmaspons:
Hi @mpadge ! I added a test. Also, I added the selected server to `packageStartupMessage()`, roxigenized the package with the changes from #428 and removed redundant _See also_ functions which where already included in the section by `@family`

mpadge (approved):

